PR #76, opened 2025-06-10
ci: add clusterfuzzlite to the repo

malto101:


chinmayajha:
been sitting for over a year and the repo has no fuzz targets, so the workflow has nothing to actually run. happy to bring this back when we add real fuzz harnesses.

PR #129, opened 2025-07-16
refactor: change from decorator to functional call

malto101:


davidamo9 (approved):

malto101:
@thakur-patel this was 1 small change i had thought earlier,
can you look into this if this is viable or not,
close otherwise

chinmayajha:
this already happened on main, action_keyword.py calls _ai_self_heal() directly now instead of going through the decorator. closing as done.

PR #172, opened 2025-08-27
chore: indexing on google search

malto101:


chinmayajha:
no movement in a year and nobody ever registered this token in search console, so the file on its own does nothing. takes seconds to re-add whenever we actually want docs indexing.

PR #262, opened 2026-03-06
feat(appium): improve deeplink handling with Android ADB fallback and…

sheshnath1st:
… iOS Safari support

### DIFF
--- a/optics_framework/engines/drivers/appium.py
+++ b/optics_framework/engines/drivers/appium.py
@@ -1086,7 +1086,7 @@ class Appium(DriverInterface):
             return None
 
     def _is_deeplink(self, value: str) -> bool:
-        return isinstance(value, str) and "://" in value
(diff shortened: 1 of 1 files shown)


optics_framework/engines/drivers/appium.py
```diff
@@ -1086,7 +1086,7 @@ def appium_find_element(self, element: str) -> Optional[Any]:
             return None
 
     def _is_deeplink(self, value: str) -> bool:
-        return isinstance(value, str) and "://" in value
+        return isinstance(value, str) and "://" in value.strip()
```
thakur-patel: is this required?

thakur-patel (changes requested):

chinmayajha:
landed on main already, appium.py has _open_android_deeplink with mobile: deepLink plus the adb am start fallback and an ios path too. closing as superseded.

PR #266, opened 2026-03-11
Fix FlowControl condition logic and Playwright text locator handling

ShreehariMenon:
Fixed FlowControl module condition evaluation to correctly return module execution results and handle else scenarios.

Improved Playwright locator normalization to support plain visible text selectors, enabling scroll_until_element_appears to work reliably.

All unit and feature tests are now passing successfully.


thakur-patel:
@ShreehariMenon could you resolve all the code comments by CoPilot AI before we can go ahead with a merge?
@malto101 a review from you would help here

optics_framework/engines/drivers/playwright.py
```diff
@@ -214,6 +225,14 @@ def press_keycode(self, keycode: str, event_name=None):
         # Use mapped key or the keycode string directly (Playwright accepts key names)
         key = key_map.get(keycode, keycode)
         run_async(self.page.keyboard.press(key))
+        # wait for youtube search results page load
```
malto101: can we remove indication of YouTube in drivers

optics_framework/engines/drivers/playwright.py
```diff
@@ -308,9 +327,10 @@ async def _swipe_element_async(self, element: str, direction: str, swipe_length:
 
 
     def scroll(self, direction: str = "down", pixels: int = 120, event_name=None):
-        for _ in range(2):
-            run_async(self.page.mouse.wheel(0, pixels if direction == "down" else -pixels))
-            run_async(self.page.wait_for_timeout(120))
+        scroll_multiplier = int(self.config.get("scroll_multiplier", 1))
```
malto101: can all the config fetch like navigation timeout and scroll multiplier be done on init, so new contributors can have an idea on existing configs for the driver

malto101: would be even better, if its also part of the docs

optics_framework/api/action_keyword.py
```diff
@@ -360,12 +360,16 @@ def swipe_until_element_appears(self, element: str, direction: str, timeout: str
         self._save_screenshot_if_available(screenshot_np, "swipe_until_element_appears")
         start_time = time.time()
         while time.time() - start_time < int(timeout):
-            result = self.verifier.assert_presence(
-                element, timeout_str="3", rule="any")
-            if result:
-                break
+            try:
```
malto101, line 363: could you tell the reason for changing this ?

optics_framework/api/flow_control.py
```diff
@@ -237,7 +237,7 @@ def _parse_variable_names(self, variables: Tuple[str, ...]) -> List[str]:
             if var_name.startswith("${") and var_name.endswith("}"):
                 var_name = var_name[2:-1].strip()
             else:
-                internal_logger.warning(
+                internal_logger.exception(
```
malto101: this is fine as a warning,
as its not breaking the framework

optics_framework/api/flow_control.py
malto101, line 348: This change will mostly need to be removed as it changes the existing semantics.

There are 3 parts involved here:
1. Inverted flag (`!`)
2. Condition statement/module
3. Target statement/module

The current behavior is:

If not inverted:
```
if condition_true:
    execute target
else:
    do nothing
```
If inverted:
```
if condition_true:
    do nothing
else:
    execute target
```
In the new implementation, after the inversion logic is applied, we effectively end up with:
```
if condition_true:
    return result if result else execute target
else:
    execute target
```
This means the target module is executed when the condition evaluates to false, which is not the current behavior for non-inverted conditions.

Additionally, when `condition_true` is true and the condition module returns a non-empty result, the target module is not executed at all and the condition result is returned instead.

So this is not a refactor of the existing logic; it changes the behavior of module conditions significantly.

malto101:
rest all looks good

optics_framework/api/action_keyword.py
ShreehariMenon: This change was added to prevent transient assert_presence() failures from terminating the retry loop immediately during scroll/swipe operations while the page content is still loading or updating.

Updated the implementation to catch only AssertionError so unexpected exceptions still surface normally.

malto101: @thakur-patel I am a bit confused on this one,
could you share your inputs?

PR #267, opened 2026-03-14
Refactor Playwright scroll behavior and normalize API

ShreehariMenon:
Refactored Playwright scroll implementation to ensure pixel parameter is respected and removed hardcoded repeated wheel events.
This improves predictability, performance and aligns driver behaviour with higher-level retry logic in action keywords.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,3 +180,5 @@ cython_debug/
 *.json
 *.log
 .DS_Store
+MagicMock/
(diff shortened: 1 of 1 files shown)


chinmayajha:
the actual diff here is two gitignore lines, no scroll changes to review against the title. closing. if the pixel param scroll work is still pending a focused PR would be welcome.

PR #273, opened 2026-04-03
APP-314 fix. Workflow now highlights asserts, presses, text fields etc as intended

chinmayajha:
Fix to https://mozarkai-orchestrator.atlassian.net/browse/APP-314

Workflow already handled annotations, but the problem was in saving of the image. Commit fixes invalid timestamp format causing screenshot save failure (on windows)

changes file saving behaviour, fixes workflow

### DIFF
--- a/optics_framework/common/utils.py
+++ b/optics_framework/common/utils.py
@@ -175,11 +175,10 @@ def parse_text_only_prefix(element: str) -> Tuple[str, bool]:
 
 def get_timestamp():
     try:
-        current_utc_time = datetime.now(timezone.utc)
-        desired_timezone = timezone(timedelta(hours=5, minutes=30))
-        current_time_in_desired_timezone = current_utc_time.astimezone(desired_timezone)
-        formatted_time = current_time_in_desired_timezone.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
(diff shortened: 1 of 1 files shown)


optics_framework/common/utils.py
```diff
@@ -273,19 +272,23 @@ def save_screenshot(img, name, output_dir, time_stamp=None):
...
-        cv2.imwrite(screenshot_file_path, img)
-        internal_logger.debug(f'Screenshot saved as : {time_stamp}-{name}.jpg')
-        internal_logger.debug(f"Screenshot saved to :{screenshot_file_path}")
+        success = cv2.imwrite(screenshot_file_path, img)
 
+        if success:
+            internal_logger.debug(f"Screenshot saved as : {time_stamp}-{name}.jpg")
+            internal_logger.debug(f"Screenshot saved to :{screenshot_file_path}")
+        else:
+            internal_logger.error(f"cv2 returned FALSE. FAILED to save screenshot to : {screenshot_file_path}")
```
thakur-patel: this error will not give context to a developer new to this repo, we can improve it

thakur-patel (changes requested):
Where is the change to highlight elements? I only see a timestamp filename change done in this PR

PR #278, opened 2026-04-17
Fixed android app-version detection and added iOS app-version support

chinmayajha:


thakur-patel (approved):

thakur-patel:
Approved this for a beta tag release. This is to be rigorously tested in UAT.

PR #285, opened 2026-05-19
fix + feat: expose on-screen error detection to library consumers + review fixes

chinmayajha:
## Summary

Expose on-screen error detection to library consumers and address review feedback.

### Feature

- Refactors the existing CLI-only on-screen error detection into a shared module (`optics_framework/common/error_detection.py`) so it can be reused by `TestRunner` (CLI) and the `Optics` class (library).
- Adds two new public methods on `Optics`:
  - `add_error_definitions(source)` — accepts a raw dict or an `ErrorDefinitions` model; merges into the active session.
  - `capture_and_detect(context_label, test_case)` — captures page source via `StrategyManager`, runs detection, returns matched dicts. Pure data primitive; driver errors are swallowed so a flaky session returns `[]` instead of raising.
- Adds `docs/usage/error_detection.md` documenting CLI CSV, library API, and matching semantics.

### Review fixes (this commit)

- **Rename `pattern` → `match_string`:** the column name now reflects that matching is a case-insensitive substring check, not regex. Updated everywhere: sample CSV, models, reader, parser, optics.py, JUnit handler, logs, and docstrings.
- **Fix regex false positives:** the malformed-XML fallback's `\b` boundary partially matched `data-text`/`aria-label`/`tab-name` as `text`/`label`/`name`. Use `(?<![\w-])` lookbehind to match only whole attribute names.
- **Narrow broad except:** limit BeautifulSoup parsing exception to `(ValueError, TypeError)` only; the ImportError is guarded at module load, so it is never masked. Log the parse failure; return empty string instead of raw HTML (avoids CSS/class false positives).
- **Add case-insensitive-substring semantics to docs:** prominent note in docstrings and new `error_detection.md` page (wired into mkdocs nav).
- **Update poetry.lock:** added beautifulsoup4 + soupsieve.
- **Merge origin/main:** resolve version conflict (1.8.7-beta.1) and bring in 38 commits from main.

## Testing

- Pre-commit (ruff, bandit, trim-whitespace, etc.) passes on all changed files.
- Import smoke test on all modified modules ✓
- End-to-end detection flow (XML extraction → substring match → result dict) ✓
- CSV header identification and loading with new `match_string` column ✓


malto101:
as its a new feature, it would be nice to have it documented when you have free time.

UPD: Added

optics_framework/optics.py
```diff
@@ -516,6 +517,100 @@ def add_module(self, module_name: str, module_def: Any) -> None:
...
+        """Load user-defined error patterns onto the active session.
+
+        Accepts a CSV path (with columns ``error_code, pattern, description,
+        severity``), a raw dict ``{code: {pattern, description, severity}}``,
+        or an :class:`ErrorDefinitions` model. Repeated calls merge into the
+        existing set (later additions overwrite same-code earlier entries).
+
+        Required before :meth:`capture_and_detect` can return any matches.
+        """
+        from optics_framework.common.runner.data_reader import CSVDataReader
```
malto101: this is supposed to be a lib, the user decides how to pass data, just expose it as API instead of passing a csv file

malto101 (changes requested):

optics_framework/common/Junit_eventhandler.py
```diff
@@ -250,6 +240,24 @@ def _update_testcase_status(self, testcase: ET.Element, event: Event, testsuite:
             testsuite.set("skipped", str(
                 int(testsuite.get("skipped", "0")) + 1))
 
+    def add_detected_errors(self, session_id: str, detected: list) -> None:
+        """Insert a <properties> block with detected on-screen errors into the session testsuite."""
```
thakur-patel: @malto101 I want your suggestions on this too. How should we define errors in the junit xml report by optics?

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+
+Used by both the CLI/TestRunner path (`_capture_end_of_run_artifacts`) and the
+library `Optics` class (`capture_and_detect`) so the matching logic lives in
+one place.
+"""
+import re
+from typing import Dict, List, Optional
+
+# Attributes in Appium XML that carry user-visible text on screen.
+_APPIUM_TEXT_ATTRS = re.compile(
```
thakur-patel: modify the function and variable names here if we are making this specific to appium only

optics_framework/common/Junit_eventhandler.py
```diff
@@ -170,17 +169,13 @@ def _handle_test_case_event(self, event: Event, session_suite: ET.Element, sessi
             elapsed = event_time - self.start_times.get(event.entity_id, event_time)
             testcase.set("time", f"{elapsed:.2f}")
             self._update_testcase_status(testcase, event, session_suite)
-            # Attach all collected keywords as children
-            for kw_element in self.keyword_elements.get(event.entity_id, []):
```
thakur-patel, line 174: why are these removed?

thakur-patel (changes requested):

optics_framework/common/runner/test_runnner.py
```diff
@@ -733,6 +807,7 @@ async def run_all(self) -> None:
             await self.execute_test_case(current.name)
             current = current.next
         self.result_printer.stop_live()
+        self._capture_end_of_run_artifacts()
```
thakur-patel, line 810: @malto101 needed your input here, is there any python native way to run this? it seems to be a tearDown function

optics_framework/optics.py
chinmayajha: this is actually how modules.csv and tests.csv work right? this file is supposed to be adjacent to them in behaviour. is it not currently?

malto101: not quite right, 
users have 3 choices
cli -> if users want to use csv(this is fine if its simple)
library -> if users wants more control or wants to integrate it other products (most user will not deal with csvs here, they would want python types to present so can have more dynamic control on it)
API -> if users want to run over a network

this is also a reason invoke_api present in library is not directly api.yaml file and instead letting users construct it.

optics_framework/common/Junit_eventhandler.py
malto101: @chinmayajha any reason you are not using <error>?

.gitignore
```diff
@@ -174,7 +174,6 @@ cython_debug/
 # PyPI configuration file
 .pypirc
 
-*.csv
```
malto101, line 177: can we return this,
if its due to not being able to add error_defination.csv
you can just use 
```
git add -f sample/contact/test_data/error_degfination.csv
```

malto101 (changes requested):

optics_framework/common/error_detection.py
malto101: I would suggest to not add anything appium specific in common folder
if we are, can we at least add placeholders for other attributes

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+    For Appium XML: pulls values of the attributes that carry on-screen text
+    (text, content-desc, label, value, hint, name), ignoring resource-ids,
+    class names, bounds, and other metadata. This prevents false-positive
+    matches where a pattern appears inside an element attribute name or CSS
+    class rather than as visible text.
+    """
+    stripped = page_source.lstrip()
+    if stripped.lower().startswith(("<html", "<!doctype")):
+        try:
+            from bs4 import BeautifulSoup
```
kun-codes: I feel this import should be on top of this file

kun-codes: beautifulsoup4 should be added to `pyproject.toml`

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+            return BeautifulSoup(page_source, "lxml").get_text(separator=" ", strip=True)
+        except Exception:
```
kun-codes: this `Exception` is too broad, can you make it more specific? It would have suppressed the `ImportError` from beautifulsoup4 too

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+            return page_source
```
kun-codes: sending back raw html can lead to false positives as errors would be searched across the html and not just the visible text

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+Used by both the CLI/TestRunner path (`_capture_end_of_run_artifacts`) and the
+library `Optics` class (`capture_and_detect`) so the matching logic lives in
+one place.
+"""
+import re
+from typing import Dict, List, Optional
+
+# Attributes in Appium XML that carry user-visible text on screen.
+_APPIUM_TEXT_ATTRS = re.compile(
+    r'\b(?:text|content-desc|label|value|hint|name)="([^"]*)"'
```
kun-codes: I tried this regex here: https://regex101.com/r/y8IlS9/1

and it has some false positives

optics_framework/common/runner/test_runnner.py
```diff
@@ -475,6 +484,71 @@ async def _handle_fallback_exhausted(self, keyword_node, module_node, keyword_re
...
+
+    def _run_end_of_run_detection(
+        self, page_source: Optional[str], timestamp: Optional[str], output_dir: str
+    ) -> None:
+        try:
+            error_defs = getattr(self.session, "error_definitions", None)
+            if not (error_defs and error_defs.get_all() and page_source):
+                return
+            searchable = extract_visible_text(page_source)
+            matched = detect_errors_in_text(searchable, error_defs.get_all())
```
kun-codes: This line and https://github.com/mozarkai/optics-framework/pull/285/changes#diff-1e3b87b6b4fc75a7120144ef1412e15a99e6521d4b724c92a1c53605f63acabeR604


these 2 lines pass different things into `detect_errors_in_text()` which would lead to extra keys: `detected_at` and `test_case`

I am not sure if this is a bug but this looks like an inconsistency

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,70 @@
...
+
+    `matched_on` is ``"pattern"`` when the pattern column triggered the match,
+    or ``"code"`` when only the error_code substring was present.
+    """
+    matched: List[Dict] = []
+    for code, meta in error_definitions.items():
+        pattern = meta.get("pattern", "")
+        matched_on: Optional[str] = None
+        if pattern and pattern in searchable:
+            matched_on = "pattern"
```
kun-codes: the name "pattern" seems misleading as user may think that it refers to regex but internally it is being searched using substring matching

kun-codes: this is case sensitive matching, does the user know about this?

kun-codes (changes requested):

chinmayajha:
@thakur-patel @kun-codes  I have added fixes for the changes requested in the comments. Please review the latest commit.

optics_framework/common/error_detection.py
kun-codes: resolved

kun-codes: @chinmayajha run a `poetry lock` and commit the lock file too

optics_framework/common/error_detection.py
kun-codes: not really resolved, I will add in an updated comment for it

optics_framework/common/error_detection.py
kun-codes: will this pr also get its own docs like https://github.com/mozarkai/optics-framework/tree/1cf707faa8d309a15943e3230bb4779f7161290e/docs if so then the case sensitive matching should also be written in the markdown files along with the comment

optics_framework/common/error_detection.py
kun-codes: not resolved yet

optics_framework/common/error_detection.py
kun-codes: not resolved yet

optics_framework/common/error_detection.py
kun-codes: not yet resolved

optics_framework/common/error_detection.py
```diff
@@ -0,0 +1,99 @@
...
+    For Appium / mobile XML: parses via ElementTree and collects values of
+    the attributes that carry on-screen text (text, content-desc, label,
+    value, hint, name), ignoring resource-ids, class names, bounds, and
+    other metadata.  Falls back to regex extraction when the XML is
+    malformed.
+    """
+    stripped = page_source.lstrip()
+    if stripped.lower().startswith(("<html", "<!doctype")):
+        if not _BS4_AVAILABLE or _BeautifulSoup is None:
+            return ""
```
kun-codes, line 57: add in a log saying that html page source error detection has been skipped and installing BeautifulSoup4 will enable it

kun-codes (changes requested):

optics_framework/common/error_detection.py
chinmayajha: added

optics_framework/optics.py
chinmayajha: ok, now it accepts a dict or ErrorDefinitions; CSV stays in the CLI/runner layer. Tightened the docstrings.

optics_framework/common/error_detection.py
chinmayajha: Replaced \b with (?<![\w-]) lookbehind. Verified against your exact regex101 string - now only text=/content-desc= match, not the look-alikes.

optics_framework/common/error_detection.py
chinmayajha: bs4 is in pyproject; ran poetry lock (adds only beautifulsoup4 + soupsieve) and committed it.

optics_framework/common/error_detection.py
chinmayajha: Narrowed to (ValueError, TypeError) around parsing only; the import is guarded at module load, so it's never masked. Added a warning log.

optics_framework/common/error_detection.py
chinmayajha: Returns "" (with a warning) when bs4 is missing or parsing fails - raw HTML is never returned.

optics_framework/common/runner/test_runnner.py
chinmayajha: not a bug. detected_at/test_case are intentionally optional context keys; the core fields (error_code, matched_on, match_string, …) are always present. Now explicitly documented in the function's docstring.

optics_framework/common/error_detection.py
chinmayajha: Renamed to "match_string" everywhere: sample CSV header, ErrorDefinitions model, CSVDataReader, execute.py header-sniffing, optics.py, JUnit handler, test-runner log, and the matched_on value.

chinmayajha: > this is case sensitive matching, does the user know about this?

It's already case-insensitive substring matching; documented prominently in docstrings and the new docs page.

> will this pr also get its own docs like https://github.com/mozarkai/optics-framework/tree/1cf707faa8d309a15943e3230bb4779f7161290e/docs if so then the case sensitive matching should also be written in the markdown files along with the comment

yes, added

optics_framework/common/Junit_eventhandler.py
chinmayajha: added

PR #289, opened 2026-05-26
Fix: CSS element type detection

thakur-patel:
As per CSS spec, a valid #id must start with a letter, underscore, or hyphen after #.
Strings that begin with a digit after # are not valid CSS selectors and will now fall through to Text. E.g. #91 will be identified as Text.

### DIFF
--- a/optics_framework/common/utils.py
+++ b/optics_framework/common/utils.py
@@ -133,40 +133,31 @@ def escape_csv_value(s: str) -> str:
 
 
 def determine_element_type(element):
+    element = element.strip()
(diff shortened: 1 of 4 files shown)


thakur-patel:
@malto101, do you have any idea why the pre-commit check isn't running here?

malto101:
> @malto101, do you have any idea why the pre-commit check isn't running here?

its done, GitHub has some bug which makes the runner to hang,
if it occurs in future, just close and reopen the PR

Cheers!

malto101 (approved):

PR #291, opened 2026-05-29
feat: add 'optics live' interactive keyword session command

chinmayajha:
## Introduce `optics live` — an interactive, driver-agnostic keyword session

Adds `optics live`: a full-screen terminal session for building tests keyword-by-keyword against a live target, recording them as you go, and saving the result as a reusable module. It works across every optics driver and includes an optional natural-language mode.

### What it does
- **Interactive TUI** (alternate-screen, like lazygit): type a keyword call, run it against the device/browser, and see pass/fail, timing, and the winning locator strategy inline. Keyword/element autocomplete, ghost-text parameter hints, a `Ctrl-K` keyword browser, and per-session logs.
- **Record & save**: every successful action is buffered in order; `/save <name>` writes a `modules/` + `test_cases/` CSV and snapshots the run's screenshots/artifacts. Other commands: `/device /elements /screenshot /help /quit`.

### Driver-agnostic & config-driven
- The driver comes entirely from the project's `config.yaml`, so the same flow works for **Appium (Android/iOS), Selenium, and Playwright** — web, mobile, TV.
- Requires a `config.yaml` with **exactly one** enabled `driver_source` + an enabled `elements_source`. The status bar labels the target by driver (e.g. `playwright:chromium`, `appium:emulator-5554`). Device discovery/switching is gated to Android/Appium only (iOS/web never invoke `adb`), and config parse/validation errors are surfaced clearly.

### Natural-language mode (`Ctrl-N`, optional)
- Toggle AI mode and type plain English (*"type 'movies for kids' in the search bar"*); a bounded ReAct loop screenshots the screen, asks Gemini for the next keyword, executes it, and repeats until done/failed/aborted (`Ctrl-X`). A fully successful run is recorded and `/save`-able.
- Backed by a **reusable LLM engine type** (`llm_models` config block → `LLMFactory` → `GeminiLLM` implementing `LLMInterface`), the same shape as the driver/OCR/image engines, so the integration can be reused beyond live.
- The `google-genai` SDK is an **optional extra** (`pip install 'optics-framework[llm]'`) and is only imported when enabled — non-AI users are unaffected. Setup is documented in `docs/usage/live_usage.md`.

### Also included in this PR
- `fix(flow-control)`: module-based conditions now evaluate as true on a non-empty result (per the documented contract) and return the condition module's own result.

### Testing & docs
- Unit suites for the live controller (config validation, per-driver targets, Android-gated switching), the NL agent, and the LLM wiring; full unit suite green.
- `docs/usage/live_usage.md` covers usage, the config requirement, samples (Appium/Selenium/Playwright), and natural-language setup.

----------------------

### Deferred work
- [ ] **`flow_control` module-condition fix** — pulled out of this PR; ship as a dedicated PR with a deprecation/compat path and validation against real scripts. (Two tests `test_condition_module_*_no_else` are `xfail`'d as placeholders.)
- [ ] **NL screenshot resilience** — retry capture + a short settle (wire `halt_duration`) so a transient black-frame `E0303` doesn't kill a whole AI run. (Planned, not implemented.)
- [ ] **`E0401` false-negative** — gate `_find_error_log` on "no winning strategy recorded" so a self-healed (OCR) success isn't reported as a failure, removing the redundant `text_only:` retry the LLM does today. (Intentionally left as-is.)

### Architecture / tech-debt
- [ ] **Move device discovery behind `DriverInterface`** — keep `adb`/`idevice_id` knowledge out of the `optics live` layer (reviewer note; kept as-is for now).
- [ ] **TUI library consolidation** — standardize on one library (Textual) and drop `prompt_toolkit`, per the base-deps-size suggestion.
- [ ] **Cross-platform `/device` switch** — auto-reconfigure platform caps when switching to a device of a different platform (currently fails gracefully).

### DIFF
--- a/docs/usage/live_usage.md
+++ b/docs/usage/live_usage.md
@@ -1,7 +1,7 @@
 # Live Usage (`optics live`)
 
 `optics live` opens a full-screen, interactive terminal session for building tests
-keyword-by-keyword against a real device. It looks and behaves like Claude Code or
+keyword-by-keyword against a live target (device or browser). It looks and behaves like Claude Code or
 lazygit: the screen is taken over and redrawn in place, with a persistent input box
 and status bar pinned at the bottom. Every successful action is recorded so you can
 save the session as a reusable module.
@@ -9,25 +9,32 @@ save the session as a reusable module.
 ## Launch
 
 ```bash
-optics live                  # zero-config: works against the first connected device
-optics live <project_folder> # use your project's config + elements
+optics live                  # uses the config.yaml in the current directory
+optics live <project_folder> # uses the config.yaml in that project folder
 ```
 
-The folder is optional. With no config (or no folder), Optics auto-detects the
-first Android device reported by `adb` and opens an Appium session using minimal
-capabilities (`platformName: Android`, `automationName: UiAutomator2`, default
-server at `http://127.0.0.1:4723`) — no `appPackage` required. You drop straight
-into whatever screen the device is on (home screen, current app) and can `swipe`,
-`press_keycode`, etc.
+`optics live` is **driver-agnostic and config-driven**: the same flow works for
+Android/iOS (Appium), web (Selenium, Playwright), TV, etc. The driver comes entirely
+from the project's `config.yaml`, so a session "just works" for whatever you configure.
 
-When a config.yaml **is** present, anything you specify wins and only missing
-sections fall back to defaults. Named elements from the project are loaded lazily
-the first time they are needed.
+A `config.yaml` is **required** — there are no built-in defaults. It must declare:
 
-The session is opened automatically on launch (a `launch_app` action appears as
-the first history entry). If your project's config specifies an `appPackage` /
-`appActivity`, that app is launched; otherwise the session attaches without
-starting any app.
+- **exactly one** enabled entry under `driver_sources` (`appium`, `selenium`,
+  `playwright`, …), and
+- at least one enabled entry under `elements_sources`.
+
+If no usable config is found, or the config has no enabled driver / more than one
+enabled driver / no enabled element source, `optics live` exits with a clear message.
+A malformed `config.yaml` reports the actual parse/validation error (not "no config").
+
+See `optics_framework/samples/` for ready-made configs: `contact` (Appium/Android),
+`gmail_web` (Selenium), `playwright` (Playwright). Named elements from the project are
+loaded lazily the first time they are needed.
+
+The session is opened automatically on launch (a `launch_app` action appears as the
+first history entry). Each driver interprets `launch_app` using its own config: Appium
+launches the configured `appPackage`/`appActivity`, while Selenium/Playwright navigate
+to the configured URL.
 
 ## Layout
 
@@ -36,8 +43,9 @@ starting any app.
   winning locator strategy (`[XPath]`, `[Text]`, `[OCR]`, `[Image]`). New entries are
   appended and the view auto-scrolls to the newest.
 - **Input box** (pinned): where you type keyword calls and slash commands.
-- **Status bar** (pinned, bottom): active device, the always-on recording indicator
-  (`rec ●`), and a hint of available commands.
+- **Status bar** (pinned, bottom): the active **target** — labelled by driver, e.g.
+  `appium:emulator-5554`, `selenium:chrome`, `playwright:chromium` — the always-on
+  recording indicator (`rec ●`), and a hint of available commands.
 
 ## Running keywords
 
@@ -73,7 +81,7 @@ sleep 5
 | Command          | Description |
 |------------------|-------------|
 | `/save <name>`   | Save the recorded actions to `modules/<name>.csv` + `test_cases/<name>.csv`, **and** snapshot every screenshot/artifact the framework generated this session to `execution_output/<name>/`. Re-saving updates the snapshot. |
-| `/device [id]`   | List connected devices; with no argument, pick one from a list to switch the active device (single active device only). |
+| `/device [id]`   | **Android/Appium sessions only.** List connected `adb` devices; with no argument, pick one to switch the active device. For Selenium/Playwright/iOS sessions it reports that switching doesn't apply (the target is the configured browser/device). |
 | `/elements`      | Open a read-only popup of named elements and their locators (Esc closes). |
 | `/screenshot`    | Capture the current device screen to a file and note the path in the history. |
 | `/help`          | Show the command reference (Esc closes). |
--- a/docs/usage/live_usage.md
+++ b/docs/usage/live_usage.md
@@ -81,7 +81,7 @@ sleep 5
 | Command          | Description |
 |------------------|-------------|
 | `/save <name>`   | Save the recorded actions to `modules/<name>.csv` + `test_cases/<name>.csv`, **and** snapshot every screenshot/artifact the framework generated this session to `execution_output/<name>/`. Re-saving updates the snapshot. |
-| `/device [id]`   | **Android/Appium sessions only.** List connected `adb` devices; with no argument, pick one to switch the active device. For Selenium/Playwright/iOS sessions it reports that switching doesn't apply (the target is the configured browser/device). |
+| `/device [id]`   | **Appium sessions only.** List all connected **Android** (`adb`) and **iOS** (`idevice_id`) devices, each labelled by platform; with no argument, pick one to switch the active device's `udid`. The chosen device must match the session's configured platform. For Selenium/Playwright it reports that switching doesn't apply (the target is the configured browser). |
 | `/elements`      | Open a read-only popup of named elements and their locators (Esc closes). |
 | `/screenshot`    | Capture the current device screen to a file and note the path in the history. |
 | `/help`          | Show the command reference (Esc closes). |
--- a/docs/usage/live_usage.md
+++ b/docs/usage/live_usage.md
@@ -1,0 +1,120 @@
+# Live Usage (`optics live`)
+
+`optics live` opens a full-screen, interactive terminal session for building tests
+keyword-by-keyword against a real device. It looks and behaves like Claude Code or
+lazygit: the screen is taken over and redrawn in place, with a persistent input box
+and status bar pinned at the bottom. Every successful action is recorded so you can
+save the session as a reusable module.
+
+## Launch
+
+```bash
+optics live                  # zero-config: works against the first connected device
+optics live <project_folder> # use your project's config + elements
+```
+
+The folder is optional. With no config (or no folder), Optics auto-detects the
+first Android device reported by `adb` and opens an Appium session using minimal
+capabilities (`platformName: Android`, `automationName: UiAutomator2`, default
+server at `http://127.0.0.1:4723`) — no `appPackage` required. You drop straight
+into whatever screen the device is on (home screen, current app) and can `swipe`,
+`press_keycode`, etc.
+
+When a config.yaml **is** present, anything you specify wins and only missing
+sections fall back to defaults. Named elements from the project are loaded lazily
+the first time they are needed.
+
+The session is opened automatically on launch (a `launch_app` action appears as
+the first history entry). If your project's config specifies an `appPackage` /
+`appActivity`, that app is launched; otherwise the session attaches without
+starting any app.
+
+## Layout
+
+- **History pane** (top, scrollable): one entry per executed action showing the call
+  as typed, pass/fail status (`✓` / `✗` / `⋯` while running), execution time, and the
+  winning locator strategy (`[XPath]`, `[Text]`, `[OCR]`, `[Image]`). New entries are
+  appended and the view auto-scrolls to the newest.
+- **Input box** (pinned): where you type keyword calls and slash commands.
+- **Status bar** (pinned, bottom): active device, the always-on recording indicator
+  (`rec ●`), and a hint of available commands.
+
+## Running keywords
+
+Type a keyword call and press Enter, for example:
+
+```
+launch_app
+press_element ${login_btn} index=0
+enter_text ${username} "hello world"
+sleep 5
+```
+
+- Keyword names come live from the framework's `KeywordRegistry`, so autocomplete
+  always matches what the runner supports.
+- `${name}` references resolve against the project's named elements, with the same
+  fallback behaviour as the batch runner (each locator is tried in order).
+- `key=value` tokens are passed as keyword arguments.
+- A failing keyword is shown as `✗` with a short error (and error code) and is **not**
+  recorded; the prompt returns ready for the next command. The UI never crashes.
+
+## Autocomplete & hints
+
+- **Keyword completion** — start typing the first token and press Tab.
+- **Element completion** — type `${` to get element names, each shown with its first
+  locator.
+- **Ghost-text parameter hints** — once a keyword is recognised, its parameter
+  signature is shown dimmed after the cursor: required params in `<>`, optional in `[]`.
+- **Keyword browser** — press `Ctrl-K` for a navigable list of every keyword
+  (Up/Down to move, Enter to drop it into the input box, Esc to close).
+
+## Slash commands
+
+| Command          | Description |
+|------------------|-------------|
+| `/save <name>`   | Save the recorded actions to `modules/<name>.csv` + `test_cases/<name>.csv`, **and** snapshot every screenshot/artifact the framework generated this session to `execution_output/<name>/`. Re-saving updates the snapshot. |
+| `/device [id]`   | List connected devices; with no argument, pick one from a list to switch the active device (single active device only). |
+| `/elements`      | Open a read-only popup of named elements and their locators (Esc closes). |
+| `/screenshot`    | Capture the current device screen to a file and note the path in the history. |
+| `/help`          | Show the command reference (Esc closes). |
+| `/quit`          | End the session, run the normal driver teardown/cleanup, and exit. |
+
+## Keys
+
+| Key            | Action |
+|----------------|--------|
+| `Enter`        | Run the command, or accept the highlighted completion |
+| `Tab` / `S-Tab`| Cycle completions |
+| `${`           | Suggest element names |
+| `Ctrl-K`       | Toggle the keyword browser |
+| `Esc`          | Close any popup or the keyword browser |
+| `Ctrl-C`       | Quit |
+
+## Recording & saving
+
+Recording is always on. Every successful keyword is appended to an in-memory buffer
+in the order it ran. The buffer is only written to disk when you run `/save`. If you
+`/quit` with unsaved actions, you are warned once — run `/save <name>` to keep them,
+or `/quit` again to discard and exit.
+
+### What `/save` persists vs. what `/quit` discards
+
+The framework auto-generates screenshots and other diagnostic artifacts for every
+keyword call (a pre-action screenshot, the post-action result image, AOI captures,
+etc.). During a live session those land in a temp directory so they don't litter
+your project.
+
+* `/save <name>` snapshots that temp directory to `execution_output/<name>/` —
+  so the screenshots that accompanied your recorded steps stay alongside the saved
+  module. Re-running `/save` with the same name refreshes the snapshot.
+* `/quit` (without a prior `/save`, or after discarding the unsaved-prompt) deletes
+  the temp directory. Anything you explicitly chose to persist — `/save` outputs,
+  `/screenshot` captures in `screenshots/` — is untouched.
+
+## Logs
+
+Every live session writes a chronological log of both the framework's internal
+and execution loggers to `<project>/logs/optics_live_<timestamp>.log`. The path is
+shown as the first entry in the history pane on startup, and again on stderr after
+you `/quit`. Logs survive `/quit` regardless of whether you `/save` — they're for
+diagnostics, not for the saved script.
--- a/docs/usage/live_usage.md
+++ b/docs/usage/live_usage.md
@@ -155,19 +155,23 @@ in the order it ran. The buffer is only written to disk when you run `/save`. If
 `/quit` with unsaved actions, you are warned once — run `/save <name>` to keep them,
 or `/quit` again to discard and exit.
 
-### What `/save` persists vs. what `/quit` discards
+### Screenshots are saved automatically
 
-The framework auto-generates screenshots and other diagnostic artifacts for every
-keyword call (a pre-action screenshot, the post-action result image, AOI captures,
-etc.). During a live session those land in a temp directory so they don't litter
-your project.
+The framework auto-generates screenshots for every keyword call (a pre-action
+screenshot, the post-action result image, and annotated/AOI captures). In a live
+session these are written to a **persistent per-session folder**,
+`screenshots/session_<timestamp>/`, and **survive `/quit`** — so every keyword you run
+(typed or AI-driven) leaves a visual record with no extra step. The `<timestamp>`
+matches the session log (`logs/optics_live_<timestamp>.log`) so the two correlate. An
+empty session folder (no screenshots captured) is removed on exit; otherwise it's kept.
 
-* `/save <name>` snapshots that temp directory to `execution_output/<name>/` —
-  so the screenshots that accompanied your recorded steps stay alongside the saved
-  module. Re-running `/save` with the same name refreshes the snapshot.
-* `/quit` (without a prior `/save`, or after discarding the unsaved-prompt) deletes
-  the temp directory. Anything you explicitly chose to persist — `/save` outputs,
-  `/screenshot` captures in `screenshots/` — is untouched.
+`/save <name>` additionally bundles a **copy** of that session's screenshots into
+`execution_output/<name>/`, alongside the saved module, so a saved test is
+self-contained. Re-running `/save` with the same name refreshes the copy.
+
+> The screenshots that the AI mode sends to the model are captured separately and are
+> not written to disk; what you see in `screenshots/session_<timestamp>/` are the
+> keyword pre/post/annotated frames.
 
 ## Logs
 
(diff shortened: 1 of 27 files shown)


optics_framework/helper/live_tui.py
```diff
@@ -0,0 +1,701 @@
...
+    HSplit,
+    Window,
+    WindowAlign,
+)
+from prompt_toolkit.layout.controls import BufferControl, FormattedTextControl
+from prompt_toolkit.layout.dimension import Dimension
+from prompt_toolkit.layout.menus import CompletionsMenu
+from prompt_toolkit.layout.processors import AfterInput
+from prompt_toolkit.styles import Style
+from prompt_toolkit.widgets import Frame
```
malto101, line 36: i would suggest to use either prompt_toolkit or textual which is already being used to reduce dependencies. Current optics with base deps(no drivers) is coming to ~500M

P.S. i personally use textual, because you can use mouse clicks in TUI

ref: [config_manager](https://github.com/mozarkai/optics-framework/blob/c30389c176664960eef5cc8775ff0acdfa73838b/optics_framework/helper/config_manager.py#L1-L7)

malto101 (commented):
Great work, just look into the deps as a chore

optics_framework/helper/live_tui.py
chinmayajha: I would have done that, but there are trade-offs
- The inline autocomplete dropdown (keyword names + ${element} completion).
- The "ghost text" inline parameter hints (press_element <element> [repeat]).
- Inline completion-while-typing behavior.
would have to let go of these

malto101: See, either side works
Either remove textual, or prompt toolkit

chinmayajha: I explored `optics setup`
Migrating the old one to the new one. Not doing it in this PR as it's not related. Will keep it as an open PR for contribution.

^ This is because features are intact, and priority wise migration can be done a bit later.

thanks for reviewing

chinmayajha:
@thakur-patel @malto101 
Updated commits and PR description

ARYANK-08 (approved):
LGTM - reviewed system prompt

optics_framework/common/llm_interface.py
```diff
@@ -0,0 +1,72 @@
...
+        temperature: Optional[float] = None,
+    ) -> Dict[str, Any]:
+        """
+        Convenience wrapper: call :meth:`generate` in JSON mode and parse the result.
+
+        Concrete (non-abstract) so every engine inherits consistent JSON handling.
+
+        :raises OpticsError: ``Code.E0801`` if the model returns undecodable JSON.
+        """
+        import json
```
thakur-patel: can this import be at the top? json is a light import, so it should be okay

optics_framework/api/flow_control.py
```diff
@@ -346,37 +346,58 @@ def _is_module_condition(self, cond_str: str) -> Optional[bool]:
         return self.modules.get_module_definition(actual_cond) is not None
 
     def _handle_module_condition(self, cond_str: str, target: str) -> Optional[List[Any]]:
-        """Handles evaluation and execution for module-based conditions."""
```
thakur-patel, line 349: Can you add the old vs new semantics for this if it modified?

thakur-patel (changes requested):

docs/usage/live_usage.md
```diff
@@ -0,0 +1,184 @@
...
+screenshot **and** a condensed UI hierarchy of the on-screen elements (their text,
+content-desc, resource ids, bounds, and flags — when the driver exposes a page source),
+and drives keywords step-by-step until the goal is reached; `Ctrl-X` aborts. A fully
+successful run is recorded and `/save`-able like any manual session. It works with whatever
+driver your config uses (Appium/Selenium/Playwright). Three steps to enable it:
+
+**1. Install the optional LLM extra** (the SDK is not in the base install):
+
+```bash
+pip install 'optics-framework[llm]'
```
thakur-patel, line 116: also, add the llm driver definitions to [setup.py](https://github.com/mozarkai/optics-framework/blob/main/optics_framework/helper/setup.py)

optics_framework/common/nl_agent.py
```diff
@@ -0,0 +1,366 @@
...
+
+            terminal = self._run_one_step(instruction, catalog, catalog_names, state, on_step)
+            if terminal is not None:
+                return terminal
+
+        return AgentResult(
+            "exhausted", state.history, "Reached the maximum number of steps.", state.successful
+        )
+
+    # -- helpers ---------------------------------------------------------------
```
thakur-patel: let's clean up comments like this one

optics_framework/engines/llm_models/gemini.py
```diff
@@ -0,0 +1,106 @@
...
+        system: Optional[str] = None,
+        response_schema: Optional[Dict[str, Any]] = None,
+        *,
+        temperature: Optional[float] = None,
+    ) -> str:
+        contents: List[Any] = [prompt]
+        if images:
+            for img in images:
+                contents.append(
+                    genai_types.Part.from_bytes(data=img, mime_type="image/png")
```
thakur-patel: is there a specific reason to only accepting png images?

optics_framework/engines/llm_models/gemini.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Google Gemini engine (``google-genai`` SDK), supporting both the Gemini Developer API
+    and Vertex AI. With no explicit ``capabilities`` the SDK auto-selects the backend and
+    credentials from the environment (``GOOGLE_GENAI_USE_VERTEXAI``,
+    ``GOOGLE_API_KEY`` / ``GEMINI_API_KEY``, ``GOOGLE_CLOUD_PROJECT``,
+    ``GOOGLE_CLOUD_LOCATION``, ``GOOGLE_APPLICATION_CREDENTIALS``). Capabilities override env
+    when present; secrets are never hardcoded.
+    """
+
+    NAME = "gemini"
```
thakur-patel: I don't see NAME used anywhere. Does it have any future usage planned?

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+                msg += "  — run `launch_app` first to open the session."
+            return msg
+        return f"{type(exc).__name__}: {exc}"
+
+    # -- Recording / save ---------------------------------------------------------
+
+    def save(self, name: str) -> Tuple[str, str, Optional[str]]:
+        """Persist the recorded actions and their accompanying artifacts.
+
+        Writes:
```
thakur-patel, line 735: Where do we save test data like - elements? 
I'm thinking about a user's first test case creation journey. He won't have any saved elements then.

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+
+    def _build_candidates(self, params: List[str]) -> List[List[str]]:
+        """Expand each positional ``${element}`` into its list of fallback locators."""
+        candidates: List[List[str]] = []
+        for param in params:
+            if param.startswith("${") and param.endswith("}"):
+                var_name = param[2:-1].strip()
+                values = self.session.elements.get_element(var_name) if self.session.elements else None
+                if not values:
+                    raise OpticsError(Code.E0201, message=f"Element not found: {var_name}")
```
thakur-patel: I presume this is where we are trying to get more locators for an element. 
"E0201: Element not found" isn't the ideal error message here. I suggest to create a new error message group for optics live for new cases like these

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+        for record in capture.records:
+            if isinstance(record, logging.LogRecord) and record.levelno >= logging.ERROR:
+                try:
+                    return record.getMessage()
+                except Exception:  # pragma: no cover - defensive
+                    return str(record)
+        return None
+
+    @staticmethod
+    def _winning_strategy(capture: LogCaptureBuffer) -> Optional[str]:
```
thakur-patel: capture as a variable name isn't very implicit for a LogCaptureBuffer type. 
I think we can use some better naming here - logCapture or something would make better sense.
Same goes for _find_error_log() too

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+    def _enabled_driver_name(self) -> str:
+        """Name of the enabled driver source (e.g. ``appium``/``selenium``/``playwright``).
+
+        ``_compose_config`` guarantees exactly one is enabled; ``unknown`` is a defensive
+        fallback that should not occur in practice.
+        """
+        drivers = _enabled_drivers(self.config)
+        return drivers[0] if drivers else "unknown"
+
+    def _target_from_config(self) -> Optional[str]:
```
thakur-patel: _get_target_id_from_config() would be a more implicit name for this

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+        elif self.driver_type == "ble":
+            return caps.get("port") or caps.get("device_id")
+        return None
+
+    def active_target(self) -> str:
+        """Human-readable label of the live target for the status bar.
+
+        e.g. ``appium:emulator-5554``, ``playwright:chromium``, ``selenium:chrome``.
+        """
+        label = self.active_target_label or self._target_from_config()
```
thakur-patel: why is the OR required when the definition is already done in the constructor? does that fail anytime?

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+        if label:
+            return f"{self.driver_type}:{label}"
+        return self.driver_type or "unknown"
+
+    def supports_device_switching(self) -> bool:
```
thakur-patel, line 843: add a todo or create a task on github to add support for iOS devices too
It's not a scope of this PR so its okay, but create a todo so it doesn't get forgotten

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+        require ``platformName == android``. Missing/unknown platform -> False, so we
+        never run adb against an iOS or non-mobile session.
+        """
+        if self.driver_type != "appium":
+            return False
+        caps = self._enabled_driver_caps() or {}
+        return str(caps.get("platformName", "")).lower() == "android"
+
+    @staticmethod
+    def list_devices() -> List[str]:
```
thakur-patel: Do a platform check inside to keep the android code and ios code; you can use idevice_id to get a list of ios device ids.
The other way round is to rename the function so that it doesn't confuses developers who missed reading the comments. 
I would strongly suggest to go with the former approach. I can help in testing.

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+        """Switch the active device by rebuilding the driver/session with new capabilities.
+
+        Android/Appium only — device hot-swap via adb serial has no meaning for web
+        (selenium/playwright) or iOS sessions.
+        """
+        if not self.supports_device_switching():
+            raise OpticsError(
+                Code.E0501,
+                message=(
+                    f"Device switching is for Android/Appium only; this session uses "
```
thakur-patel: Good job giving the clear error here

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+            raise OpticsError(Code.E0303, message="Screenshot capture unavailable")
+        import cv2
+
+        image = self._action_keyword.strategy_manager.capture_screenshot()
+        ok, buffer = cv2.imencode(".png", image)
+        if not ok:  # pragma: no cover - defensive
+            raise OpticsError(Code.E0303, message="Failed to encode screenshot")
+        return buffer.tobytes()
+
+    # -- Natural-language mode ----------------------------------------------------
```
thakur-patel, line 987: if possible, create separate files because this file is getting too long

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+    text: str
+    result: Optional[ActionResult] = None  # set when kind == "keyword"
+
+
+@dataclass
+class NLSummary:
+    """Terminal summary of a natural-language run."""
+
+    instruction: str
+    status: str  # PASS | FAIL | ABORTED | MAX_STEPS
```
thakur-patel: Create an Enum for these kind of values

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+    # history pane can render it as an indented child / muted thinking line.
+    nl_child: bool = False
+    nl_thinking: bool = False
+
+
+@dataclass
+class NLStep:
+    """A single streamed event from a natural-language run, for the history pane."""
+
+    kind: str  # "thinking" | "keyword" | "note"
```
thakur-patel: Create an Enum for these kind of values

optics_framework/helper/live.py
```diff
@@ -0,0 +1,1246 @@
...
+
+
+@dataclass
+class ActionResult:
+    """Outcome of a single keyword execution, ready to render in the history pane."""
+
+    raw: str
+    keyword: str = ""
+    params: List[str] = field(default_factory=list)
+    status: str = "PASS"  # PASS | FAIL | INFO | RUNNING
```
thakur-patel: Create an Enum for these kind of values

thakur-patel (changes requested):

optics_framework/common/llm_interface.py
chinmayajha: changed

optics_framework/helper/live.py
chinmayajha: Elements is something they have to provide themselves. `live` only creates test cases and modules.csv

optics_framework/helper/live.py
chinmayajha: added

optics_framework/helper/live.py
chinmayajha: added

optics_framework/helper/live.py
chinmayajha: added

optics_framework/helper/live.py
chinmayajha: refactoring commons/ and big files is a separate (non-prriority) PR.

optics_framework/common/nl_agent.py
chinmayajha: done, made them bit more nuanced wherever needed, and deleted wherever not needed

docs/usage/live_usage.md
chinmayajha: Added an "LLM Drivers" category → Gemini → google-genai

optics_framework/api/flow_control.py
chinmayajha: documented it now

old: "didn't raise" -> true, returned target's result
new: non-empty-result -> true, returns condition's result

optics_framework/engines/llm_models/gemini.py
chinmayajha: changed, sniffs mime from magic bytes (PNG/JPEG/GIF/WebP), defaults PNG — no longer tied to one encoding

optics_framework/engines/llm_models/gemini.py
chinmayajha: removed. was part of an idea for multiple LLM support but has been scrapped (for now)

optics_framework/helper/live.py
chinmayajha: Added Category.LIVE + E0901 "Named element not defined…"; _build_candidates/_resolve_value now raise it instead of reusing E0201

optics_framework/helper/live.py
chinmayajha: Renamed → log_capture in _find_error_log/_winning_strategy

optics_framework/helper/live.py
chinmayajha: renamed

optics_framework/helper/live.py
chinmayajha: added TODO

optics_framework/helper/live.py
chinmayajha: committed this addition, please test

optics_framework/helper/live.py
chinmayajha: yes it is redundant, removed

optics_framework/api/flow_control.py
thakur-patel: The comment in code added is too bulky, can you make it short and direct? 
More importantly, whenever we do a semantic change, we have to ensure backwards compatibility. Existing scripts using optics will break once this is released, so do have a look at the impact of this change. I'd suggest to abstract backwards compatibility into a separate function that can be removed easily later without the initial author's intervention. 
Also, update the necessary optics-framework docs if it isn't yet updated. I see that you have updated the code based on the documented contract, just do a double-check post current usage validation from an existing scripter

optics_framework/helper/live.py
thakur-patel: Let's follow this process for separate PRs :- 

The GitHub-Native Approach (Best for Visibility)

1. Create a Tracked Issue Immediately: Don't wait until the current PR is merged. Create the new issue now.
2. Use GitHub Task Lists: In your current PR description, add a markdown checklist that links to the new issue (or just states the task). GitHub will show a progress bar (e.g., 0 of 1 tasks) on the PR conversation page, making it obvious that something is outstanding.

Markdown would look like this:- 
### Post-Merge Tasks
- [ ] #1245 (Create separate PR for config migration)

Convert Draft PRs Early: If you already know what code changes are needed, branch off your current branch, make the quick change, and open a Draft PR. Title it something obvious like [HOLD] Follow-up to PR #XYZ.

optics_framework/helper/live.py
thakur-patel: I understand that, but this got me thinking as to how would a user author scripts in that case. 
We need to launch an Element Extractor GUI too where optics-framework will provide a list of elements of the current page. Users can select and get that saved to elements.
The test creation journey through optics only will be incomplete without this.

thakur-patel: We can have that in a separate PR, but just create a new issue for it, a GitHub task list, or better, a draft PR

optics_framework/api/flow_control.py
chinmayajha: Noted. Pulling this fix from this PR. I thought this would be a simple hotfix but this deserves it own PR.

malto101:
Hello @chinmayajha , could you rebase your commits to be better categorized
there seems to be a lot of refactor/merge/fix commits which could be a part of your dev branch but not final PR branch
its quite a lot of info overhead as I dont get flow on where to start reviewing
many features like ai self heal and flowcontrol contract failure seems totally separate, can we have new PRs for these two, these can act as a dep for this PR to get merged

optics_framework/engines/llm_models/gemini.py
```diff
@@ -0,0 +1,119 @@
...
+    The interface passes raw image bytes (callers may send PNG, JPEG, WebP, GIF);
+    we detect rather than assume so the engine isn't tied to one encoding.
+    """
+    if data[:3] == b"\xff\xd8\xff":
+        return "image/jpeg"
+    if data[:6] in (b"GIF87a", b"GIF89a"):
+        return "image/gif"
+    if data[:4] == b"RIFF" and data[8:12] == b"WEBP":
+        return "image/webp"
+    return "image/png"
```
malto101: this is hopeful assumption,
create default to fail

optics_framework/engines/llm_models/gemini.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class GeminiLLM(LLMInterface):
```
malto101, line 45: just a question?
why are we going with vendor sdk
and not orchestrator sdk like pydantic AI, crewAi etc. that should handle multiple vendor model routing, input ingestion etc for you

optics_framework/engines/llm_models/gemini.py
```diff
@@ -0,0 +1,119 @@
...
+
+        caps: Dict[str, Any] = (config or {}).get("capabilities") or {}
+        self.model_name: str = caps.get("model", _DEFAULT_MODEL)
+        self.temperature: float = caps.get("temperature", 0.0)
+
+        client_kwargs: Dict[str, Any] = {}
+        # Explicit overrides win over env vars; absent kwargs fall back to SDK env auto-config.
+        if "use_vertexai" in caps:
+            client_kwargs["vertexai"] = bool(caps["use_vertexai"])
+        api_key = caps.get("api_key") or caps.get("gemini_api_key")
```
malto101, line 75: this is a security vulnerability
please do not save api keys in config if the config is not encrypted

optics_framework/common/ai_self_heal.py
```diff
@@ -0,0 +1,280 @@
...
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+from optics_framework.common.error import OpticsError
+from optics_framework.common.llm_interface import LLMInterface
+from optics_framework.common.logging_config import internal_logger
+
+
+# Provider callables (best-effort; may return None when unavailable).
+ScreenshotProvider = Callable[[], Optional[bytes]]
+PagesourceProvider = Callable[[], Optional[str]]
```
malto101, line 26: page source will be tricky, they get quite large making you needlessly consume context, it would be better to use elements as it extracts from pagesoource mostly

optics_framework/common/config_handler.py
```diff
@@ -38,6 +39,10 @@ class Config(BaseModel):
     event_attributes_json: Optional[str] = None
     halt_duration: float = 0.1
     max_attempts: int = 3
+    # AI self-heal: when every normal locate strategy fails for a keyword, an LLM
```
malto101: unnecessary comments

optics_framework/common/error.py
```diff
@@ -82,6 +83,9 @@ class Code(str, Enum):
     E0801 = "E0801"
     E0802 = "E0802"
 
+    # Live (interactive session) issues
+    E0901 = "E0901"
```
malto101, line 87: good start, but you need to expand it in future, current usage is quite broad.

optics_framework/common/utils.py
```diff
@@ -179,12 +179,13 @@ def get_timestamp():
         internal_logger.error('Unable to get current time', exc_info=e)
         return None
 
-def encode_numpy_to_base64(image: np.ndarray) -> str:
+def encode_numpy_to_png_bytes(image: np.ndarray) -> bytes:
```
malto101, line 182: this looks good,
just know
ideally I would like a common format across
I have been maintaining b64 as that common format as it is very portable.

optics_framework/common/utils.py
```diff
@@ -311,6 +321,76 @@ def annotate_element(frame, centre_coor, bbox):
...
+    if text:
+        parts.append(f'"{text}"')
+    desc = " ".join((el.get("content-desc") or "").split())
+    if desc:
+        parts.append(f'desc="{desc}"')
+    rid = el.get("resource-id") or ""
+    if rid:
+        parts.append(f"id={rid.split('/', 1)[-1]}")  # drop package prefix
+    bounds = el.get("bounds")
+    if bounds:
```
malto101: this is very appium focussed.
very android appium focussed

could you make It more generic
or at least create placeholder NotImplemented function so you can implement it later

optics_framework/common/utils.py
```diff
@@ -311,6 +321,76 @@ def annotate_element(frame, centre_coor, bbox):
...
+    try:
+        from lxml import etree  # type: ignore
+        root = etree.fromstring(page_source.encode("utf-8"))
+    except Exception:  # noqa: BLE001 - any malformed source degrades to "no page source"
+        return ""
+
+    lines: List[str] = []
+    _walk(root, 0, lines)
+    out = "\n".join(lines)
+    if len(out) > max_chars:
```
malto101, line 438: import should always be at the top
current one is very android specific , instead a very similar function implemented is  get_interactive_elements from element sources, could we try use that

optics_framework/engines/llm_models/gemini.py
chinmayajha: That is not the primary way. This is just for people to use their own keys if they really want to through their config. Default is environment variables only.

optics_framework/common/ai_self_heal.py
chinmayajha: I have stripped down the page source, the entire page source wont be passed.

optics_framework/common/config_handler.py
chinmayajha: Fixed.

optics_framework/common/utils.py
chinmayajha: Fixed

optics_framework/common/utils.py
chinmayajha: Changed.

optics_framework/engines/llm_models/gemini.py
chinmayajha: Agreed, added as a TODO.

optics_framework/engines/llm_models/gemini.py
chinmayajha: Fixed.

chinmayajha:
PR is polluted with fixes, creating fresh PR with all changes, cleanly review-able.

optics_framework/engines/llm_models/gemini.py
malto101: could you add a warning then if they use it

optics_framework/common/ai_self_heal.py
thakur-patel: @chinmayajha can we compare the size between the new stripped down logic v/s the stripping down logic in elements? we'll go with whichever's efficient and we can simply change the elements stripping down logic if the new one's more efficient. trying to ensure a single place for doing any changes to logic,

optics_framework/engines/llm_models/gemini.py
chinmayajha: added

optics_framework/common/ai_self_heal.py
chinmayajha: strip_page_source is ~6× smaller than the elements extraction (and ~3× smaller than the raw source). The elements path is actually larger than the raw page source.

PR #293, opened 2026-06-02
feat(mcp): expose every keyword via Model Context Protocol

chinmayajha:
Add an MCP server so AI clients (Claude Desktop, Claude Code, Cursor, claude.ai web) can drive an Optics session end-to-end through the same keyword surface CSV/YAML test cases use.

- optics_framework/mcp/ : Server("optics-framework") that fans out across session / inspect / keyword capability handlers, mirroring the mozark-mcp capability pattern.
- tools/keywords.py auto-generates one MCP tool per entry in discover_keywords() and dispatches through expose_api.execute_keyword, so MCP inherits the existing fallback / event / template logic unchanged.
- Two transports: streamable HTTP (default; Starlette + raw ASGI route at /mcp to avoid the slash-redirect that drops POST bodies) and stdio for local clients.
- New `optics mcp [--transport http|stdio] [--host] [--port] [--cors-origin]` CLI subcommand.
- Unit tests cover schema generation, blocklist, dispatch wiring, and session handler plumbing.


optics_framework/mcp/server.py
```diff
@@ -0,0 +1,107 @@
...
+        ],
+    )
+
+
+@server.list_tools()
+async def list_tools() -> list[mcp_types.Tool]:
+    return ALL_TOOLS
+
+
+@server.call_tool()
```
malto101, line 95: i would suggest to list get pagesource, get screenshot, etc(anything related to fetching data on current screen) as a resource instead of tool

chinmayajha: .

chinmayajha:
Closed because a better approach using FastMCP exists. New PR soon

PR #303, opened 2026-06-15
feat(live): refactor AI self-heal to use keyword-based ReAct loop and improve locator logging

chinmayajha:
## Summary
This PR refactors the AI self-heal mechanism from coordinate-guessing (which frequently tapped blind/guessed screen coordinates and often missed targets) to a keyword-based execution model. The AI now acts as a ReAct loop over the framework's own action keywords, utilizing more reliable, targeted actions that route through the locating strategies ladder (XPath, text, OCR, image-matching).

Additionally, it cleans up noisy expected element-not-found log output from Appium element sources to prevent the live TUI silent error checker from overriding successful self-heals with failure statuses.

### 1. AI Self-Heal Refactor (`optics_framework/common/ai_self_heal.py`)
- Rewrote `AISelfHealHandler` to run a keyword-based loop (similar to the NL agent but lighter).
- Replaced the coordinate schema (`percent_x`, `percent_y`, `text`, `direction`) in `HEAL_ACTION_SCHEMA` with `keyword` name, parameter string list, and a `completed` status flag.
- Added catalog introspection to the prompt so the LLM knows available signatures.
- Instructed the LLM to prefer targeting elements by visible text (using the condensed UI hierarchy) via `press_element` instead of guessing percentages.
- Added explicit prompt guidelines for gesture directions to align the LLM's spatial logical scroll actions with the physical swipe inputs expected by the driver.

### 2. Action Keyword Integration (`optics_framework/api/action_keyword.py`)
- Implemented `_heal_execute()` to parse and execute keywords generated by the LLM during the self-heal process.
- Temporarily disables the self-heal toggle during internal keyword calls to prevent infinite recursion.
- Added `_heal_catalog()` to serve a focused keyword catalog: `press_element`, `enter_text`, `scroll`, `swipe_by_percentage`, `press_keycode`, `press_by_percentage`.

### 3. Log Cleanup for Silent Error Checker
- **`optics_framework/engines/elementsources/appium_find_element.py`**
- **`optics_framework/engines/elementsources/appium_page_source.py`**
  - Caught `NoSuchElementException` specifically and raised `OpticsError` without logging stack tracebacks via `internal_logger.exception` (which is treated as a silent error/failure in `live.py`).

### 4. Tests
- **`tests/units/common/test_ai_self_heal.py`**
  - Updated unit tests to use `FakeExecutor` instead of `FakeDriver`.
  - Added new coverage for keyword execution, multi-step navigation, catalog prompts, done/give_up loops, and `ActionKeyword` integrations.

## Verification
- Ran full test suite:
  ```bash
  poetry run pytest --ignore=tests/feature

### DIFF
--- a/optics_framework/api/action_keyword.py
+++ b/optics_framework/api/action_keyword.py
@@ -1,14 +1,18 @@
 from functools import wraps
 import collections
+import inspect
+import shlex
 import time
 import json
-from typing import Callable, Optional, Any, Tuple
+from typing import Callable, Optional, Any, Tuple, List, Dict
 from optics_framework.common.logging_config import internal_logger
 from optics_framework.common.optics_builder import OpticsBuilder
 from optics_framework.common.strategies import StrategyManager
 from optics_framework.common.base_factory import InstanceFallback
 from optics_framework.common import utils
-from optics_framework.common.ai_self_heal import AISelfHealHandler, HealContext
+from optics_framework.common.ai_self_heal import (
+    AISelfHealHandler, HealContext, HealKeywordSpec, KeywordExecResult,
+)
 from optics_framework.common.error import OpticsError, Code
 from .verifier import Verifier
 
@@ -101,7 +105,19 @@ def _try_results_until_success(
 ):
     last_exception = None
     result_count = 0
-    for result in results:
+    locate_error: Optional[OpticsError] = None
+
+    try:
+        results_list = list(results)
+    except OpticsError as e:
+        # The locate generator raises when no strategy yields a result; treat that
+        # as "no results" so self-heal below gets a chance, but keep the original
+        # error as the cause instead of losing it.
+        internal_logger.debug(f"Locate generator raised for '{element}' in '{func_name}': {e}")
+        results_list = []
+        locate_error = e
+
+    for result in results_list:
         result_count += 1
         _save_annotated_for_result(result, screenshot_np, execution_dir, func_name)
         try:
@@ -118,7 +134,11 @@ def _try_results_until_success(
     if result_count == 0:
         if self._ai_self_heal(element, func_name, args, kwargs, screenshot_np):
             return None
-        raise OpticsError(Code.E0201, message=f"No valid strategies found for '{element}' in '{func_name}'")
+        raise OpticsError(
+            Code.E0201,
+            message=f"No valid strategies found for '{element}' in '{func_name}'",
+            cause=locate_error,
+        ) from locate_error
     if last_exception:
         if self._ai_self_heal(element, func_name, args, kwargs, screenshot_np):
             return None
@@ -223,6 +243,17 @@ class ActionKeyword:
         self._ai_healer: Optional[AISelfHealHandler] = None
         # Breadcrumbs of recently-succeeded keywords, fed to the LLM as context on heal.
         self._recent_steps: "collections.deque[Tuple[str, list]]" = collections.deque(maxlen=10)
+        # The focused subset of keywords the self-healer is allowed to call, bound here
+        # (rather than looked up by name via getattr) so a rename shows up as a static
+        # attribute error instead of a silent "Unknown keyword" at runtime.
+        self._heal_dispatch: Dict[str, Callable] = {
+            "press_element": self.press_element,
+            "enter_text": self.enter_text,
+            "scroll": self.scroll,
+            "swipe_by_percentage": self.swipe_by_percentage,
+            "press_keycode": self.press_keycode,
+            "press_by_percentage": self.press_by_percentage,
+        }
 
     def _record_successful_step(self, func_name: str, element: Any, args: tuple) -> None:
         self._recent_steps.append((func_name, [str(element), *map(str, args)]))
@@ -254,7 +285,11 @@ class ActionKeyword:
             ),
         )
         if self._ai_healer is None:
-            self._ai_healer = AISelfHealHandler(self._llm, self.driver)
+            self._ai_healer = AISelfHealHandler(
+                self._llm,
+                keyword_executor=self._heal_execute,
+                keyword_catalog=self._heal_catalog,
+            )
         internal_logger.info("AI self-heal: attempting recovery for '%s' on '%s'", func_name, element)
         result = self._ai_healer.heal(ctx, providers.screenshot, providers.pagesource)
         self._log_heal_outcome(result, func_name, element, args)
@@ -273,9 +308,78 @@ class ActionKeyword:
                 "AI self-heal: could not recover '%s': %s", func_name, result.message
             )
             return
-        action = result.action.action if result.action else "?"
-        internal_logger.info("AI self-heal: recovered '%s' via '%s'", func_name, action)
+        action = result.action
+        kw = action.keyword if action else "?"
+        internal_logger.info("AI self-heal: recovered '%s' via keyword '%s'", func_name, kw)
         self._record_successful_step(func_name, element, args)
+
+    # -- Self-heal keyword executor and catalog --------------------------------
+
+    def _heal_execute(self, line: str) -> KeywordExecResult:
+        """Keyword executor for self-heal: parse and run one keyword, returning a KeywordExecResult.
+
+        Calls the ActionKeyword methods directly but avoids re-triggering self-heal by
+        passing ``located=`` for self-healing-decorated methods or calling non-decorated ones.
+        """
+        try:
+            tokens = shlex.split(line, posix=True)
+        except ValueError as exc:
+            return KeywordExecResult(ok=False, message=f"Parse error: {exc}")
+        if not tokens:
+            return KeywordExecResult(ok=False, message="Empty keyword line")
+
+        keyword_name = tokens[0]
+        params = tokens[1:]
+
+        method = self._heal_dispatch.get(keyword_name)
+        if method is None:
+            return KeywordExecResult(ok=False, message=f"Unknown keyword: {keyword_name}")
+
+        try:
+            # Non-self-healing methods: call directly with params.
+            # Self-healing methods (press_element, enter_text): they will go through
+            # `with_self_healing` decorator, but self-heal is already running, so
+            # we need to temporarily disable it to avoid recursion.
+            original_enabled = self.ai_self_heal_enabled
+            self.ai_self_heal_enabled = False
+            try:
+                method(*params)
+            finally:
+                self.ai_self_heal_enabled = original_enabled
+            return KeywordExecResult(ok=True)
+        except OpticsError as exc:
+            return KeywordExecResult(ok=False, message=str(exc.message))
+        except Exception as exc:  # noqa: BLE001 - never crash the heal loop
+            return KeywordExecResult(ok=False, message=str(exc))
+
+    def _heal_catalog(self) -> List[HealKeywordSpec]:
+        """Return the keyword catalog for the self-healer (focused subset with signatures)."""
+        return [
+            HealKeywordSpec(name=name, signature=self._heal_keyword_signature(name, method))
+            for name, method in self._heal_dispatch.items()
+        ]
+
+    @staticmethod
+    def _heal_keyword_signature(name: str, method: Callable) -> str:
+        """Build a ghost-text signature like ``press_element <element> [repeat] [offset_x]``."""
+        try:
+            sig = inspect.signature(method)
+        except (TypeError, ValueError):
+            return name
+        parts: List[str] = [name]
+        for pname, param in sig.parameters.items():
+            if pname == "self":
+                continue
+            # Skip internal keyword-only params (located, event_name, aoi_*)
+            if param.kind in (inspect.Parameter.KEYWORD_ONLY, inspect.Parameter.VAR_KEYWORD):
+                continue
+            if pname in ("event_name", "aoi_x", "aoi_y", "aoi_width", "aoi_height"):
+                continue
+            if param.default is inspect.Parameter.empty:
+                parts.append(f"<{pname}>")
+            else:
+                parts.append(f"[{pname}]")
+        return " ".join(parts)
 
     def _capture_screenshot_safe(self) -> Any:
         """Capture a screenshot, returning None on failure (e.g. secure/protected pages)."""
--- a/optics_framework/common/ai_self_heal.py
+++ b/optics_framework/common/ai_self_heal.py
@@ -2,16 +2,20 @@
 
 When the normal element-location ladder (XPath -> on-screen text -> OCR -> image) exhausts
 for a locate-based keyword, :class:`AISelfHealHandler` asks an :class:`LLMInterface` to look at
-the current screen (screenshot + condensed page source) plus the keyword's intent and recent
-context, and to take ONE corrective device action at a time (tap / type / swipe / scroll) until
-the keyword's goal is achieved or a small step budget is hit.
+the current screen (screenshot + condensed page source) plus the keyword's intent and the
+available keyword catalog, and to emit ONE keyword call at a time (e.g. ``press_element "Meesho"``,
+``scroll "down"``) until the keyword's goal is achieved or a small step budget is hit.
 
-The handler runs *independently* of :class:`StrategyManager`: it drives the device directly via
-the :class:`DriverInterface` primitives, so the LLM must FINISH THE JOB itself (it cannot defer
-back to the normal locators). It is decoupled from any controller — it depends only on an ``llm``,
-a ``driver``, and screenshot/page-source provider callables — so it is unit-testable with fakes.
+Unlike the old coordinate-guessing approach, this routes through the framework's own keyword
+methods — so ``press_element "Meesho"`` uses the full locate ladder (XPath -> text -> OCR -> image)
+instead of the LLM blindly tapping pixel percentages.
+
+The handler is decoupled from any controller — it depends only on an ``llm``, a keyword executor
+callable, a keyword catalog callable, and screenshot/page-source provider callables — so it is
+unit-testable with fakes.
 """
 
+import shlex
 import time
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, List, Optional, Tuple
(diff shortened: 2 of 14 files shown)


optics_framework/common/strategies.py
```diff
@@ -735,6 +735,28 @@ def capture_screenshot(self) -> Optional[np.ndarray]:
...
+        frame from :meth:`capture_screenshot`.
+        """
+        for strategy in self.screenshot_strategies:
+            source = strategy.element_source
+            try:
+                data = source.capture_screenshot_bytes()
+            except NotImplementedError:
+                continue
+            except Exception as e:
+                internal_logger.debug(
```
thakur-patel, line 755: should we add an execution_tracer here? it's present in the other functions.

optics_framework/common/elementsource_interface.py
```diff
@@ -24,6 +24,21 @@ def capture(self) -> numpy.ndarray:
         """
         pass
 
+    def capture_screenshot_bytes(self) -> bytes:
```
malto101, line 27: as the end result is a raw screenshot,
do you think it would be better just to merge both capture and capture_screenshot_bytes into and pass the need as a param instead?

thakur-patel (changes requested):

optics_framework/api/action_keyword.py
```diff
@@ -101,7 +103,13 @@ def _try_results_until_success(
 ):
     last_exception = None
     result_count = 0
-    for result in results:
+
+    try:
+        results_list = list(results)
+    except OpticsError:
```
kun-codes: why is it swallowing `OpticsError`? Wouldn't this make debugging harder?

optics_framework/engines/elementsources/appium_screenshot.py
```diff
@@ -47,6 +50,29 @@ def capture(self) -> np.ndarray:
...
+        :meth:`capture_screenshot_as_numpy` does, for callers that only need encoded
+        bytes. The retry absorbs the truncated-response ("Incorrect padding") base64
+        failures occasionally seen against busy remote hubs.
+        """
+        driver = self._require_driver()
+        last_exc: Optional[Exception] = None
+        for attempt in range(2):
+            try:
+                return base64.b64decode(driver.get_screenshot_as_base64())
+            except Exception as e:
```
kun-codes: This exception is too wide as well

optics_framework/engines/elementsources/appium_screenshot.py
```diff
@@ -47,6 +50,29 @@ def capture(self) -> np.ndarray:
...
+        Return the device screen as native PNG bytes via Appium's base64 endpoint.
+
+        Skips the ``base64 -> numpy -> cv2.imdecode`` decode that
+        :meth:`capture_screenshot_as_numpy` does, for callers that only need encoded
+        bytes. The retry absorbs the truncated-response ("Incorrect padding") base64
+        failures occasionally seen against busy remote hubs.
+        """
+        driver = self._require_driver()
+        last_exc: Optional[Exception] = None
+        for attempt in range(2):
```
kun-codes: I would suggest not using magic numbers like 2. Use a named constant for this

optics_framework/engines/elementsources/selenium_screenshot.py
```diff
@@ -38,25 +41,41 @@ def get_interactive_elements(self, filter_config: Optional[List[str]] = None):
...
+        Skips the ``base64 -> numpy -> cv2.imdecode`` decode that
+        :meth:`capture_screenshot_as_numpy` does, for callers that only need encoded
+        bytes. The retry absorbs transient/truncated responses from remote Grid nodes.
+        """
+        driver = self._require_driver()
+        last_exc: Optional[Exception] = None
+        for attempt in range(2):
+            try:
+                return base64.b64decode(driver.get_screenshot_as_base64())
+            except Exception as e:
```
kun-codes: `Exception` is too wide

optics_framework/engines/elementsources/selenium_screenshot.py
```diff
@@ -38,25 +41,41 @@ def get_interactive_elements(self, filter_config: Optional[List[str]] = None):
...
+        """
+        Return the screen as native PNG bytes via Selenium's base64 endpoint.
+
+        Skips the ``base64 -> numpy -> cv2.imdecode`` decode that
+        :meth:`capture_screenshot_as_numpy` does, for callers that only need encoded
+        bytes. The retry absorbs transient/truncated responses from remote Grid nodes.
+        """
+        driver = self._require_driver()
+        last_exc: Optional[Exception] = None
+        for attempt in range(2):
```
kun-codes: We should use a named constant instead of 2 here

optics_framework/common/ai_self_heal.py
```diff
@@ -200,44 +261,67 @@ def _safe_call(provider: Optional[Callable[[], Any]]) -> Any:
...
 
-        if action.action in _LAYOUT_ACTIONS:
-            # Handler bypasses keyword-level settling; let the UI finish animating.
-            time.sleep(_SETTLE_SECONDS)
-        return action.action in _TERMINAL_ACTIONS
+        line = self._build_line(action.keyword, action.params)
+        result = self.keyword_executor(line)
+
+        # Let the UI settle after the keyword runs.
+        time.sleep(_SETTLE_SECONDS)
```
kun-codes: This runs unconditionally before checking if the goal is completed. So a successful action will have to wait here. 

Only non completing keywords or failed actions should wait for 1.5 seconds

optics_framework/engines/elementsources/appium_find_element.py
```diff
@@ -112,8 +113,10 @@ def locate(self, element: str, index: Optional[int] = None) -> Any:
             try:
                 found_element = driver.find_element(AppiumBy.ACCESSIBILITY_ID, element)
                 return found_element
+            except NoSuchElementException as e:
+                raise OpticsError(Code.E0201, message=f"Element of type {element_type} not found using: {element}", cause=e) from e
             except Exception as e:
-                internal_logger.exception(f" element: {element}", exc_info=e)
+                internal_logger.debug(f"Unexpected error finding element {element}: {e}")
```
kun-codes: This should be an exception. Its for an error

optics_framework/engines/elementsources/appium_find_element.py
```diff
@@ -128,8 +131,12 @@ def locate(self, element: str, index: Optional[int] = None) -> Any:
 
                 found_element = found_elements[index_to_use]
                 return found_element
+            except OpticsError:
+                raise
+            except NoSuchElementException as e:
+                raise OpticsError(Code.E0201, message=f"Element of type {element_type} not found using: {element}", cause=e) from e
             except Exception as e:
-                internal_logger.exception(f" element: {element}", exc_info=e)
+                internal_logger.debug(f"Unexpected error finding class element {element}: {e}")
```
kun-codes: This should be an exception. Its for an error

optics_framework/engines/elementsources/appium_page_source.py
```diff
@@ -79,6 +81,36 @@ def get_interactive_elements(self, filter_config: Optional[List[str]] = None):
...
+        else:
+            xpath = self.find_xpath_from_text(element)
+        try:
+            execution_logger.debug(f"Finding element by text: {element} with xpath: {xpath}")
+            element_obj = driver.find_element(AppiumBy.XPATH, xpath)
+            return element_obj
+        except NoSuchElementException as e:
+            raise OpticsError(Code.E0201, message=f"Element of type {element_type} not found using: {element}", cause=e) from e
+        except Exception as e:
+            internal_logger.debug("Error finding element by text: %s: %s", xpath, e)
```
kun-codes: This should be an exception. Its for an error

optics_framework/engines/elementsources/appium_page_source.py
```diff
@@ -79,6 +81,36 @@ def get_interactive_elements(self, filter_config: Optional[List[str]] = None):
...
+    def _locate_by_xpath(self, driver: WebDriver, element: str, element_type: str) -> Any:
+        if self.driver is not None and hasattr(self.driver, "ui_helper") and self.driver.ui_helper is not None:
+            xpath, _ = self.driver.ui_helper.find_xpath(element)
+            try:
+                element_obj = driver.find_element(AppiumBy.XPATH, xpath)
+                return element_obj
+            except NoSuchElementException as e:
+                raise OpticsError(Code.E0201, message=f"Element of type {element_type} not found using: {element}", cause=e) from e
+            except Exception as e:
+                internal_logger.debug("Error finding element by xpath: %s: %s", xpath, e)
```
kun-codes: This should be an exception. Its for an error

optics_framework/engines/elementsources/appium_page_source.py
```diff
@@ -164,8 +176,10 @@ def locate_using_index(self, element, index, strategy=None) -> Optional[Any]:
                 xpath = self.driver.ui_helper.get_view_locator(strategy=strategy, locator=locator)
                 try:
                     element_obj = self._require_webdriver().find_element(AppiumBy.XPATH, xpath)
-                except Exception:
-                    internal_logger.exception("Error finding element by index: %s", xpath)
+                except NoSuchElementException:
+                    return None
+                except Exception as e:
+                    internal_logger.debug("Error finding element by index: %s: %s", xpath, e)
```
kun-codes: This should be an exception. Its for an error

optics_framework/engines/elementsources/selenium_screenshot.py
```diff
@@ -38,25 +41,41 @@ def get_interactive_elements(self, filter_config: Optional[List[str]] = None):
         internal_logger.exception("SeleniumScreenshot does not support getting interactive elements.")
         raise NotImplementedError("SeleniumScreenshot does not support getting interactive elements.")
 
+    def capture_screenshot_bytes(self) -> bytes:
```
kun-codes, line 40: this and `capture_screenshot_bytes()` of `appium_screenshot` look similar

This should be extracted to a common place

optics_framework/engines/elementsources/selenium_screenshot.py
```diff
@@ -1,11 +1,14 @@
...
 import base64
+import time
 import cv2
 import numpy as np
-from selenium.common.exceptions import ScreenshotException
 from optics_framework.common.elementsource_interface import ElementSourceInterface
 from optics_framework.common.logging_config import internal_logger
 
+# Delay between screenshot retries (Selenium Grid nodes can be remote/flaky too).
+_SCREENSHOT_RETRY_BACKOFF_S = 0.5
```
kun-codes: This constant and the one in `appium_screenshot.py` are duplicated

optics_framework/helper/live.py
```diff
@@ -975,12 +975,16 @@ def capture_screenshot(self) -> str:
...
+        fast path (when a backend supports it) and the numpy fallback are chosen
+        driver-agnostically.
+        """
         if self._action_keyword is None:  # pragma: no cover - defensive
             raise OpticsError(Code.E0303, message="Screenshot capture unavailable")
-        image = self._action_keyword.strategy_manager.capture_screenshot()
         try:
-            return utils.encode_numpy_to_png_bytes(image)
+            return self._action_keyword.strategy_manager.capture_screenshot_bytes()
         except ValueError as exc:  # pragma: no cover - defensive
```
kun-codes: the `capture_screenshot_bytes()` never raises a `ValueError`. It will raise a `OpticsError` with code `E303`

optics_framework/api/action_keyword.py
```diff
@@ -1,14 +1,16 @@
 from functools import wraps
```
kun-codes, line 1: Nothing wrong with the code but commits f079ea1cbb5cf3d32c7d288db6c9aaae09831a9b and 2b545fa0a17d8aed6505b808fdcaa82a65767cc9 do not use conventional commits

kun-codes (changes requested):

optics_framework/api/action_keyword.py
chinmayajha: good catch, was masking the real error. now it gets logged and kept as the cause instead of just getting dropped

optics_framework/engines/elementsources/appium_screenshot.py
chinmayajha: fair point, narrowed it to the exceptions we actually expect from the driver here

optics_framework/engines/elementsources/appium_screenshot.py
chinmayajha: done, pulled it into a named constant

optics_framework/engines/elementsources/selenium_screenshot.py
chinmayajha: yeah same deal as the appium one, narrowed the except

optics_framework/engines/elementsources/selenium_screenshot.py
chinmayajha: done, named constant now

optics_framework/common/ai_self_heal.py
chinmayajha: good point, moved the sleep so it only fires when we're actually waiting on something (failed step or one that needs another screenshot)

optics_framework/engines/elementsources/appium_find_element.py
chinmayajha: fixed, switched to logger.exception so we don't lose the traceback

optics_framework/engines/elementsources/appium_find_element.py
chinmayajha: same fix here

optics_framework/engines/elementsources/appium_page_source.py
chinmayajha: fixed

optics_framework/engines/elementsources/appium_page_source.py
chinmayajha: fixed

optics_framework/engines/elementsources/appium_page_source.py
chinmayajha: fixed

optics_framework/engines/elementsources/selenium_screenshot.py
chinmayajha: you're right, pulled the shared bit out into one helper both use now

optics_framework/engines/elementsources/selenium_screenshot.py
chinmayajha: same fix, the constant lives in one place now

optics_framework/helper/live.py
chinmayajha: that was leftover from an older version, removed it. the OpticsError with E0303 already covers this

optics_framework/api/action_keyword.py
chinmayajha: yep, cleaned that up

optics_framework/api/action_keyword.py
```diff
@@ -273,10 +306,88 @@ def _log_heal_outcome(self, result: Any, func_name: str, element: Any, args: tup
...
+
+    def _heal_execute(self, line: str) -> Any:
+        """Keyword executor for self-heal: parse and run one keyword, returning an ExecResult-like.
+
+        Calls the ActionKeyword methods directly but avoids re-triggering self-heal by
+        passing ``located=`` for self-healing-decorated methods or calling non-decorated ones.
+
+        Returns an object with ``.ok`` (bool) and ``.message`` (str).
+        """
+        from dataclasses import dataclass as _dc
```
kun-codes: lets move this import to the top of the file and not make an 'alias' for `dataclass` as it affects readability

optics_framework/api/action_keyword.py
```diff
@@ -273,10 +306,88 @@ def _log_heal_outcome(self, result: Any, func_name: str, element: Any, args: tup
...
-        action = result.action.action if result.action else "?"
-        internal_logger.info("AI self-heal: recovered '%s' via '%s'", func_name, action)
+        action = result.action
+        kw = action.keyword if action else "?"
+        internal_logger.info("AI self-heal: recovered '%s' via keyword '%s'", func_name, kw)
         self._record_successful_step(func_name, element, args)
 
+    # -- Self-heal keyword executor and catalog --------------------------------
+
+    def _heal_execute(self, line: str) -> Any:
```
kun-codes: I believe every branch of it is returning a `_Result` object, so change the return type

optics_framework/api/action_keyword.py
```diff
@@ -273,10 +306,88 @@ def _log_heal_outcome(self, result: Any, func_name: str, element: Any, args: tup
...
+
+        Calls the ActionKeyword methods directly but avoids re-triggering self-heal by
+        passing ``located=`` for self-healing-decorated methods or calling non-decorated ones.
+
+        Returns an object with ``.ok`` (bool) and ``.message`` (str).
+        """
+        from dataclasses import dataclass as _dc
+
+        @_dc
+        class _Result:
```
kun-codes: since this method is returning `_Result` and the return value is being used in other functions, lets move this class to top of the file. If the return value is also used in other files then extract `_Result` to a common area

optics_framework/common/strategies.py
```diff
@@ -735,6 +735,27 @@ def capture_screenshot(self) -> Optional[np.ndarray]:
...
+        via the default base-class implementation.
+        """
+        for strategy in self.screenshot_strategies:
+            source = strategy.element_source
+            try:
+                data = source.capture_screenshot_bytes()
+                if data:
+                    execution_tracer.log_attempt(strategy, "screenshot_bytes", "success")
+                    return data
+            except Exception as e:
```
kun-codes, line 753: We have a bare `Exception` here

kun-codes (changes requested):

optics_framework/api/action_keyword.py
chinmayajha: moved off the local alias, dropped `_Result` entirely instead.

optics_framework/api/action_keyword.py
chinmayajha: fixed, return type is now `HealResult`.

optics_framework/api/action_keyword.py
chinmayajha: turns out ai_self_heal.py already has a `HealResult` dataclass with the same ok/message shape, so reused that instead of adding a new common type.

optics_framework/common/strategies.py
chinmayajha: left this one as bare except Exception — every other strategy-fallback loop in this class does the same (screenshot, pagesource, interactive elements), since each backend can throw different things and we just want to try the next strategy. narrowing it here would make this one inconsistent with the rest.

optics_framework/api/action_keyword.py
kun-codes: I don't think using it is correct since it has an `Optional` `action` field too

chinmayajha: you're right, that was the wrong reuse — action there means something specific to the heal loop's terminal step, not a single keyword call. Added a dedicated KeywordExecResult(ok, message) in ai_self_heal.py next to the KeywordExecutor alias instead, and _heal_execute now returns that.

kun-codes: this works @thakur-patel

malto101:
Please squash your commits
@chinmayajha

kun-codes (approved):

kun-codes:
good job @chinmayajha

thakur-patel (approved):

Issue #314, opened 2026-06-18
Add optional API authentication for `optics serve`


chinmayajha:
## Type
Security / enhancement

## Context
The REST server (`optics serve`, `common/expose_api.py`) is currently **unauthenticated** with open CORS (`allow_origins=["*"]`). This is a deliberate, documented choice that suits local CLI tooling and protected internal CI networks, but it is risky as the API grows in capability (it can drive devices, run keywords, and dry-run suites).

## Proposal
Add **opt-in**, lightweight authentication:
- API key (header, e.g. `X-API-Key`) or bearer token, configured via env var / config and enforced by a FastAPI dependency or middleware.
- Disabled by default to preserve the current local-dev/CI UX; when the key/token is set, all endpoints (including `/v1/dry_run*`) require it.
- Optionally tighten the default CORS policy when auth is enabled.

## Acceptance criteria
- With auth disabled (default), behaviour is unchanged.
- With auth configured, requests without a valid credential get `401`/`403`.
- Documented in `docs/usage/REST_API_usage.md` (update the "Authentication" section, which currently states none is required).

_Suggested labels: security, enhancement_

Issue #313, opened 2026-06-18
In-memory ZIP processing for /v1/dry_run/upload


chinmayajha:
## Type
Enhancement / performance

## Context
`dry_run_upload` (`common/expose_api.py`) securely extracts uploaded ZIPs to a temp directory on disk, and `load_suite_from_folder` (`helper/execute.py`) then reads those files back from disk. This is a round-trip through the filesystem for data we already hold in memory.

## Proposal
Allow the data readers to parse CSV/YAML directly from in-memory buffers (e.g. straight from `zipfile.ZipFile`), avoiding the temp-dir write/read.
- `CSVDataReader` / `YAMLDataReader` (`common/runner/data_reader.py`) currently accept file paths only — extend them to accept file-like objects / strings.
- Add an in-memory suite-building path parallel to `load_suite_from_folder` that categorizes entries by content (reuse `identify_file_content`) without touching disk.
- Keep all existing safety limits (zip-slip is moot in-memory, but keep entry-count / uncompressed-size / ratio caps from `common/dry_run.py`).

## Benefits
Lower latency and file I/O; fewer temp dirs to clean up.

## Acceptance criteria
- Upload dry-run works without writing suite files to disk.
- Security caps (zip-bomb/entry-count) still enforced.
- Existing tests pass; add a test asserting no temp files are created for the in-memory path.

_Suggested labels: enhancement, performance_

chinmayajha:
Outdated — the dry-run REST feature never landed on main. There are no `/v1/dry_run*` routes in `expose_api.py` (only `/health`, `/v1/sessions*`, `/v1/keywords`), and no `load_suite_from_folder` in `execute.py`; the earlier dry-run endpoints were removed by the #400/#391 rework. With no upload endpoint, there is nothing to optimize in-memory. Closing.

Issue #312, opened 2026-06-18
Async job/polling architecture for /v1/dry_run (large suites)


chinmayajha:
## Type
Enhancement

## Context
The dry-run endpoints (`POST /v1/dry_run`, `POST /v1/dry_run/upload` in `common/expose_api.py`) execute **synchronously** — the HTTP connection is held open until the entire suite is loaded and validated. For very large suites this can exceed reverse-proxy / CI gateway timeouts (commonly 30–60s).

## Proposal
Introduce an async job model for dry runs:
- `POST /v1/dry_run` (and `/upload`) returns **`202 Accepted`** immediately with an `execution_id`.
- Work runs in a background task/queue.
- `GET /v1/dry_run/{execution_id}` polls status (`PENDING`/`RUNNING`/`DONE`/`FAILED`) and returns the `DryRunResponse` when complete.

## Notes / considerations
- Reuse the existing `_execute_dry_run` core; move it behind a job runner.
- Needs a result store (in-memory dict with TTL is enough for a first cut; document the single-process limitation).
- Keep the current synchronous endpoints (or a `?wait=true`) for small suites / backward compatibility.
- Consider job cancellation and a max-concurrency cap.

## Acceptance criteria
- Submitting a suite returns `202` + `execution_id` without blocking.
- Polling returns the final structured result.
- Behaviour is documented in `docs/usage/REST_API_usage.md`.

_Suggested labels: enhancement_

chinmayajha:
Outdated — same root cause as #313: the `/v1/dry_run` REST endpoints never landed on current main (removed in the #400/#391 rework; `expose_api.py` has no dry-run routes). There is no synchronous dry-run endpoint left to make async. Closing.

Issue #311, opened 2026-06-18
Migrate Pydantic V1 `class Config` to V2 `ConfigDict`


chinmayajha:
## Type
Tech debt

## Context
The test suite emits repeated Pydantic deprecation warnings:

> `PydanticDeprecatedSince20: Support for class-based config is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0.`

These are V1-style `class Config:` blocks inside V2 `BaseModel` subclasses. They will break on the eventual Pydantic V3 upgrade.

## Scope
Migrate `class Config:` → `model_config = ConfigDict(...)` across the framework models, including (non-exhaustive):
- `DependencyConfig` — `common/config_handler.py:11`
- `Config` — `common/config_handler.py:22`
- `ModuleRegistry` — `common/base_factory.py:15`
- `InstanceFallback` — `common/base_factory.py:207`

Grep for remaining occurrences: `grep -rn "class Config:" optics_framework/`.

## Acceptance criteria
- No `PydanticDeprecatedSince20` warnings during `poetry run pytest`.
- All existing tests pass unchanged.

_Suggested labels: tech-debt_

PR #315, opened 2026-06-18
feat(api): expose optics dry_run over REST (inline + upload)

chinmayajha:
Adds two REST endpoints to validate a whole test suite without a device or Appium session:

- `POST /v1/dry_run` — inline JSON suite (`test_cases` / `modules` / `elements` / `api`)
- `POST /v1/dry_run/upload` — CSV/YAML files or a single `.zip`

Both return a per-test-case → module → keyword PASS/FAIL report, useful for catching unknown keywords, unresolved `${vars}`, and broken modules before anything gets provisioned.

## How it works

Each request builds an **ephemeral, device-less session** (new `require_driver` flag, default `True` so execute/live are unaffected) and runs it through the existing `ExecutionEngine` in dry-run mode, tearing down in `finally`. Uploaded driver configs are always stripped first, so a dry run can never connect to a device. Suite loading (`load_suite_from_folder` / `build_suite_from_inline`) is now shared between the CLI and REST paths instead of being inlined in `BaseRunner`.

## Also included

- Dry-run is now a real validator on the CLI too — an unknown keyword or unresolved `${var}` comes back as a `FAIL` with a reason instead of crashing.
- Upload hardening: zip-slip/zip-bomb guards, streaming size caps, filename sanitisation.
- Status codes: `200` ran (PASS/FAIL in body), `400` no/invalid suite, `413` too large, `422` malformed JSON.
- `LoadedSuite.config` properly typed (no more `Any`), config is copied rather than mutated before stripping sources, `DryRunResponse.status` is a `Literal`, concurrent dry-runs are capped to bound resource use.

## Testing

317 tests passing, pre-commit clean. Rebased onto latest `main`.

## Follow-ups

Async job/polling for large suites (#312), in-memory zip parsing (#313), optional API auth (#314), Pydantic v2 migration (#311).


chinmayajha:
Follow-up work intentionally left out of this PR (filed separately):

- #311 — Migrate Pydantic V1 `class Config` → V2 `ConfigDict`
- #312 — Async job/polling for `/v1/dry_run` (large suites, avoid gateway timeouts)
- #313 — In-memory ZIP parsing (skip the temp-dir round-trip)
- #314 — Optional API authentication for `optics serve`

This PR also closes out the work from #309.

malto101:
what is the use of dry run for API?

chinmayajha:
> what is the use of dry run for API?

@malto101 it's just optics dry run exposed via API.

malto101:
> > what is the use of dry run for API?
> 
> @malto101 it's just optics dry run exposed via API.

ideally API side isnt meant to run full fledge test cases due to a bunch of reason, due to that reason we also havnt added support for CRUD test cases/modules

so having dry run wouldnt make much sense here,
whole idea of dry run is to validate no code syntax by making it go through to the parser and not actually performing the driver action

chinmayajha:
@malto101 This code addition is a good-to-have.  We were aware of a business requirement at `mozark.ai` which would require the dry run exposed API to be ingested at their end for a workflow so I thought it would be a good idea to add it here. Optics APIs are ingested at multiple places, so I thought this would be a good-to-have here. 

I don't think any principle's are being violated here. please let me know if there's any other reason for not finding this feature a good addition.

malto101:
> @malto101 This code addition is a good-to-have. We were aware of a business requirement at `mozark.ai` which would require the dry run exposed API to be ingested at their end for a workflow so I thought it would be a good idea to add it here. Optics APIs are ingested at multiple places, so I thought this would be a good-to-have here.
> 
> I don't think any principle's are being violated here. please let me know if there's any other reason for not finding this feature a good addition.

with this dry run, what do plan to validate?

chinmayajha:
> with this dry run, what do plan to validate?

Structure of test suites. Doesn't have to be complete test suites. @malto101

PR #319, opened 2026-06-25
Refactor/use prompt toolkit for optics config

kun-codes:
This PR replaces use of `textual` with `prompt_toolkit`. The reason for using `prompt_toolkit` is because the size of it and its dependencies are far lower than that of `textual`

I would be posting sizes later as the internet here is currently very slow.

Please merge this pr after merging #318

### DIFF
--- a/optics_framework/common/config_handler.py
+++ b/optics_framework/common/config_handler.py
@@ -119,11 +119,9 @@ class ConfigHandler:
         "llm_models"
     ]
 
-    def __init__(self, config: Optional[Config] = None):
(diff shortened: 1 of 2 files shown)


kun-codes:
This pr is not ready for merge yet. I have to test it more

malto101:
it seems this migration even migrated the bugs in it.
could you do a full manual test on it after packaging your branch as a wheel and testing it

kun-codes:
yes I will test it out. I never tested it so always kept it as a draft

PR #322, opened 2026-06-29
feat: enhance MCP server deployment instructions and add Docker support

malto101:
Add Docker support for optics mcp (HTTP transport on port 8090)

- Containerized MCP uses HTTP only (http://<host>:8090/mcp); stdio remains for local MCP clients.
- MCP and optics serve use separate in-memory sessions — clients must call start_session per process.
- Prod MCP needs a PyPI release that includes the [mcp] extra; until then, use mcp-dev with a local wheel (poetry build + --build-arg WHL_FILE=...).

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -45,7 +45,52 @@ services:
       timeout: 10s
       retries: 5
 
+  mcp:
+    build:
+      context: ..
+      dockerfile: Docker/mcp/prod/Dockerfile
+    image: optics/mcp:prod
+    container_name: optics_mcp_prod
+    ports:
+      - "8090:8090"
+    restart: unless-stopped
+    environment:
+      - PYTHONUNBUFFERED=1
+    healthcheck:
+      test: ["CMD-SHELL", "python -c \"import socket; s=socket.create_connection(('127.0.0.1',8090),2); s.close()\""]
+      interval: 45s
+      timeout: 10s
+      retries: 5
+    deploy:
+      resources:
+        limits:
+          cpus: '1.0'
+          memory: 1024M
+
+  mcp-dev:
+    build:
+      context: ..
+      dockerfile: Docker/mcp/dev/Dockerfile
+      args:
+        # pass WHL_FILE if you want to install from a specific wheel in dist/
+        WHL_FILE: ""
+    image: optics/mcp:dev
+    container_name: optics_mcp_dev
+    ports:
+      - "8091:8090" # maps host 8091 -> container 8090 so prod MCP / mock OCR don't conflict
+    volumes:
+      - ..:/app:cached
+    environment:
+      - PYTHONUNBUFFERED=1
+    healthcheck:
+      test: ["CMD-SHELL", "python -c \"import socket; s=socket.create_connection(('127.0.0.1',8090),2); s.close()\""]
+      interval: 45s
+      timeout: 10s
+      retries: 5
+
 # Usage examples:
 #  docker compose -f Docker/docker-compose.yml up --build app
 #  docker compose -f Docker/docker-compose.yml up --build dev
+#  docker compose -f Docker/docker-compose.yml up --build mcp
+#  docker compose -f Docker/docker-compose.yml up --build mcp-dev
 #  docker compose -f Docker/docker-compose.yml up --build
--- a/Docker/mcp/dev/Dockerfile
+++ b/Docker/mcp/dev/Dockerfile
@@ -9,13 +9,12 @@ WORKDIR /app
 
 # Accept the .whl filename as a build argument
 ARG WHL_FILE
+ARG VISION_BACKEND=easyocr
+
+ENV VISION_BACKEND=${VISION_BACKEND}
 ENV WHL_FILE=${WHL_FILE}
 
-# Build arguments for vision backend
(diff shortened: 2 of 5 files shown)


Docker/mcp/dev/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+   libglib2.0-0 \
+   && rm -rf /var/lib/apt/lists/*
+
+# Install common Python packages and optics-framework wheel with MCP extra and vision backend
+RUN pip install --no-cache-dir \
+   appium-python-client \
+   && pip install --no-cache-dir "/app/${WHL_FILE}[mcp]" \
+   && if [ "$VISION_BACKEND" = "easyocr" ]; then \
+      pip install --no-cache-dir easyocr && \
+      python3 -c "import easyocr; easyocr.Reader(['en'], download_enabled=True)"; \
```
chinmayajha: in both dev and prod this will download as `root` right? but you switch to appuser later on. So the caching will be useless in that case as we will download it again during runtime always.

Docker/mcp/prod/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+    && playwright install-deps chromium firefox
+
+# Install optics-framework with MCP extra and vision backend (late layer for cache efficiency)
+RUN uv pip install --system "optics-framework[mcp]" \
+    && if [ "$VISION_BACKEND" = "easyocr" ]; then \
+        uv pip install --system easyocr && \
+        python3 -c "import easyocr; easyocr.Reader(['en'], download_enabled=True)"; \
+    elif [ "$VISION_BACKEND" = "google-vision" ]; then \
+        uv pip install --system google-cloud-vision; \
+    elif [ "$VISION_BACKEND" = "pytesseract" ]; then \
```
chinmayajha, line 36: pytesseract is a wrapper around the binary right? we need the apt-get install too i think.

Docker/docker-compose.yml
```diff
@@ -45,7 +45,51 @@ services:
...
+        limits:
+          cpus: '1.0'
+          memory: 1024M
+
+  mcp-dev:
+    build:
+      context: ..
+      dockerfile: Docker/mcp/dev/Dockerfile
+      args:
+        WHL_FILE: ""
```
chinmayajha, line 76: we should mention in the docs that this is compulsory, or it will fail to compile and users would have no clue

Docker/mcp/dev/Dockerfile
malto101: Docker layer caching is based on instructions and filesystem changes, so anything installed during build is already part of the image.


We’re not downloading anything at runtime, that would unnecessarily increase boot time.

Docker/mcp/prod/Dockerfile
malto101: yep, we need
looks like i forgot to add it

Docker/docker-compose.yml
malto101: it is already mentioned as part of deployment.md
[ref](https://github.com/malto101/optics-framework/blob/20e82b0bb873922a52efbe7e44ff491b3a004dda/Docker/deployment.md#development-mcp-local-whl)

malto101: just in case, i have also added a validation in the dockerfile

Docker/mcp/dev/Dockerfile
chinmayajha: I misstated, I meant that obviously image will have the downloaded models, agreed but the models are downloaded into root's EasyOCR cache during the build, but the container later runs as appuser

so won't EasyOCR look in a different cache directory and thendownload them at runtime again?

malto101: just noticed it.
fixed

kun-codes:
I feel we should use multistage docker builds because it would lead to smaller docker images and lesser attack surface

kun-codes:
Also for production dockerfiles, we should use [docker hardened images](https://www.docker.com/blog/docker-hardened-images-for-every-developer/) as they have been made free for use a few months back

Docker/mcp/dev/Dockerfile
chinmayajha: thank you

chinmayajha (approved):

malto101:
> I feel we should use multistage docker builds because it would lead to smaller docker images and lesser attack surface

Looked into multistage for prod earlier, it works, but only saves ~180–230 MB (3–6%) because ~6.5 GB of our image is Playwright + PyTorch/EasyOCR, which can't be build-stage-only. It also adds complexity around browser install, EasyOCR model paths, and cross-stage site-packages copies, I've had this break before. Modest compiler removal isn't worth the maintenance cost in this PR

malto101:
> Also for production dockerfiles, we should use [docker hardened images](https://www.docker.com/blog/docker-hardened-images-for-every-developer/) as they have been made free for use a few months back

we can do this once with Mozark's  account, DHI requires signing in dhi.io 

also its not a drop in replacement
as 
1. we use a lot of system packages here. dhi base doesnt have system packages off the shelf (so no apt-get)
2. a lot of frame buffer/GUI libs will be missing which is needed for our x11 server and playwright

we could you hardern the docker base image tags to SHA, to pin it in the future.

PR #331, opened 2026-07-03
fix(serve): enable file logging for HTTP API sessions via `file_log`config field

kun-codes:
Before this PR, no log files were being written to using `optics serve`. 
After this PR, file logging can be enabled using `"file_log": true,` in request body for `POST {{baseUrl}}/v1/sessions/start`

### DIFF
--- a/optics_framework/common/expose_api.py
+++ b/optics_framework/common/expose_api.py
@@ -257,6 +257,7 @@ class SessionConfig(BaseModel):
     appium_url: Optional[str] = None
     appium_config: Optional[Dict[str, Any]] = None
     api_data: Optional[Dict[str, Any]] = None  # Inline API definitions only; file path not supported in REST
+    file_log: bool = False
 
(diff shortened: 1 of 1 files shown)


malto101:
how to do you plan to fetch the file logs from containers?

kun-codes:
@malto101 How about using docker volume mounts to store the logs from the container on the host system? when `file_log` is set to true, they will get stored on the host system.

And when file logging isn't required, logs go to `stderr` which can be read through `docker logs` command

malto101:
> @malto101 How about using docker volume mounts to store the logs from the container on the host system? when `file_log` is set to true, they will get stored on the host system.
> 
> And when file logging isn't required, logs go to `stderr` which can be read through `docker logs` command

That works, just double check if there's a export logs option in docker or other container services so we do not duplicate work

kun-codes:
@malto101 `docker logs <container_name> > logs.txt 2>&1` can be used to export logs in docker. For podman the command is `podman logs <container_name> > logs.txt 2>&1`

malto101:
> @malto101 `docker logs <container_name> > logs.txt 2>&1` can be used to export logs in docker. For podman the command is `podman logs <container_name> > logs.txt 2>&1`

right,
if a soln existing, we can use it,

PR #333, opened 2026-07-06
fix: remove non standard logs

kun-codes:


malto101 (approved):
LGTM

malto101:
Damn
PR no 333

This calls for a celebration 
@thakur-patel pls sponsor

PR #330, opened 2026-07-01
Fix/fix multiple keywords

kun-codes:
A fix for bugged/not implemented keywords documented at: https://github.com/kun-codes/optics-framework/tree/docs/live-keyword-notes

Broken / doesn't work:
~~- [ ] `assert_presence`~~ (false positive)
- [x] `capture_pagesource`
- [x] `capture_screenshot`
- [x] `enter_number`
- [x] `enter_text_using_keyboard`
- [x] `evaluate`
- [ ] `get_app_version` [^1]
- [x] `get_text`
- [ ] `invoke_api`
~~- [ ] `launch_app`~~ (not supposed to be called when after session is started, cannot test since `optics serve` starts a session)

Works but has a bug:
- [x] `clear_element_text`
- [x] `close_and_terminate_app` (possibly wrong documentation about what it does)
- [ ] `date_evaluate` (cannot fix it since it needs `optics live` to work for testing before )
- [x] `enter_text`
- [x] `get_interactive_elements`
- [x] `press_by_coordinates`
- [x] `press_keycode`
- [x] `scroll_until_element_appears`
- [x] `swipe_until_element_appears`

Missing from docs:
- [x] `assert_equality`(not implemented too)
- [x] `get_screen_elements`
- [x] `is_element` (not implemented too)
- [x] `swipe_by_percentage` (wrong keyword mentioned in docs)
- [x] `select_dropdown_option` (has a bug too)

Not yet tested:
- [ ] `condition`
- [ ] `execute_module`
- [ ] `run_loop`

Additional Notes:
- `get_screen_elements` returns a screenshot in base64 through the response json which makes the json too big. It would be more efficient to return a url link through the response json which the client can download if needed
- `get_interactive_elements` without any filter returns all interactive elements which is misleading
- `close_and_terminate_app`'s name seems to be misleading. Refer to what it actually does at commit: 93a101787912a537468417c0159895d869b9344f

Additional Bugs:
- [x] Some keywords' logs are duplicated twice in log files

[^1]: uses `adb` on android. `adb` cannot reach the device since appium won't forward adb in an insecure way without configuring it on the server

PR #343, opened 2026-07-09
Fix/optics serve var resolution

kun-codes:
- needs the branch in PR https://github.com/mozarkai/optics-framework/pull/330 for verification of fixes in this branch so merge this into branch of https://github.com/mozarkai/optics-framework/pull/330 to test it locally
- merge this to main before merging pr #330

### DIFF
--- a/optics_framework/common/execution.py
+++ b/optics_framework/common/execution.py
@@ -13,7 +13,7 @@ from optics_framework.common.runner.test_runnner import TestRunner, PytestRunner
 from optics_framework.common.logging_config import LoggerContext, internal_logger
 from optics_framework.common.models import TestCaseNode
 from optics_framework.common.error import OpticsError, Code
-from optics_framework.common.utils import _is_list_type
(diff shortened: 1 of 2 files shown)


malto101:
Hello @kun-codes 
could you explain the PR?

what is it solving?
how are you fixing?

kun-codes:
@malto101 Currently when passing a variable like `${var}` to a keyword in `optics serve`, it doesn't get resolved to the value of the variable. Instead it is passed literally like "${var}". This resolution of variables is present for the csv execution and `optics live`. My pr adds it to `optics serve` too

malto101:
> @malto101 Currently when passing a variable like `${var}` to a keyword in `optics serve`, it doesn't get resolved to the value of the variable. Instead it is passed literally like "${var}". This resolution of variables is present for the csv execution and `optics live`. My pr adds it to `optics serve` too

Got it!.
does optics serve support CRUD elements/Modules/TC

PR #350, opened 2026-07-13
Feat/run tests on pr

kun-codes:
- I have used `ubuntu-latest` without any version pin since it runs python code and doesn't depend on os specific features
- I used latest major versions of the actions instead of exact hashes as done in https://github.com/mozarkai/optics-framework/blob/a966b569115b1a3684730af5721dae67a62a2cc6/.github/workflows/precommit_main.yml
- no 3rd party action has been used for installing poetry. I used the instructions given at https://python-poetry.org/docs/#installing-with-pipx

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,5 +26,8 @@ jobs:
       - name: Install dependencies
         run: poetry install --no-interaction --with dev,test
 
+      - name: Run `optics setup`
+        run: poetry run optics setup --install "EasyOCR"
+
       - name: Run tests
         run: poetry run pytest -m "not live"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
         run: poetry install --no-interaction --with dev,test
 
       - name: Run `optics setup`
-        run: poetry run optics setup --install "EasyOCR"
+        run: poetry run optics setup --install "EasyOCR" "Appium"
 
       - name: Run tests
         run: poetry run pytest -m "not live"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
         run: poetry install --no-interaction --with dev,test
 
       - name: Run `optics setup`
-        run: poetry run optics setup --install "EasyOCR" "Appium"
+        run: poetry run optics setup --install "EasyOCR" "Appium" "Playwright"
 
       - name: Run tests
         run: poetry run pytest -m "not live"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,30 @@
+name: tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version-file: 'pyproject.toml'
+
+      - name: Install Poetry
+        run: pipx install poetry
+
+      - name: Install dependencies
+        run: poetry install --no-interaction --with dev,test
+
+      - name: Run tests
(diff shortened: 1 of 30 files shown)


malto101:
Thanks for the support here,
its been lacking for quite some time

as you have a base here, its a good start but it will be quite problematic to just push the test directly, 
you will need to take care of few things
1. do you plan to run all tests or tests targeting the files in PR? (as we are using the free tier, we dont have unlimited GitHub action time limit).
2. need to make sure all tests are passing currently.
3. what is the code coverage (this is main one) -> hopefully publish the report either in docs or sonarqube on every release.
4.  Test plan(without a test plan its going to be very hard to validate features)(going to be a headache in seeing what all features does optics provide in a glance) -> have it either in docs or wiki/projects in GitHub. 


I will still suggest to version pin as it would flag as part of secops
why is it problematic to run playwright?, we can have it run in no-gui mode/ maybe see how playwright/appium is doing their testing


ref: 
https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/test-coverage/overview
https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/test-coverage/python-test-coverage

kun-codes:
1. We will target all tests because github actions for public repositories is free and unlimited. It takes around 4.5 minutes to run all the tests in the current configuration of the pr. [Here](https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions) are the docs
2. But bug fixes are not the scope of the pr. However currently it is failing due to lack of easyocr which I can fix using a optics command
3. I will have to look into this
4. I am not sure what a test plan is
5. Will using version pinned using the version tag (major.minor.patch) be okay or do I need to use commit hashes? Commit hashes look cryptic as they lack context
6. I will do it in headless mode

malto101:
> 1. We will target all tests because github actions for public repositories is free and unlimited. It takes around 4.5 minutes to run all the tests in the current configuration of the pr. [Here](https://docs.github.com/en/billing/concepts/product-billing/github-actions#free-use-of-github-actions) are the docs

got it

> 2. But bug fixes are not the scope of the pr. However currently it is failing due to lack of easyocr which I can fix using a optics command

that's true, just keeping the point here as going forward after this PR is merged, no other PR will get merged util the tests are fixed

> 3. I will have to look into this

Do look into code coverage as it will be helpful for any software project.

https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/test-coverage/overview
https://docs.sonarsource.com/sonarqube-cloud/analyzing-source-code/test-coverage/python-test-coverage

as we use sonarqube for static analysis, we will need to integrate with our existing sonarqube CI.

> 4. I am not sure what a test plan is

hmm, honestly this would be more towards self learn
the basic idea is, we dont randomly create tests, we create a plan to make sure regression do not occur in future due introductions of new code changes.
there are few ways to go through it

- Requirement-based testing: Start from the feature requirements and identify what needs to be validated. Every requirement should have at least one corresponding test case
- Regression-focused testing: Identify existing functionality that could be impacted by the new change and ensure those scenarios are validated before release
- End-to-end scenario testing – Think about how a user actually uses the application and validate complete workflows rather than isolated components

these are just few ways, The reason i am asking you to this is because, a plan is essentially a document (or sometimes just a structured list) that answers:

1. What are we testing?
2. Why are we testing it?
3. What areas could be impacted?
4. What test scenarios and test cases will be executed?

Since you are starting out, I would recommend learning how to:

1. Read the current list of features of optics and understand them, then convert them to test scenarios
2. Document your test approach before developing tests

> 5. Will using version pinned using the version tag (major.minor.patch) be okay or do I need to use commit hashes? Commit hashes look cryptic as they lack context

you will need to use commit hashes, as version tags can be compromised quite easily.
if you feel it lack context, you can add a comment. 

> 6. I will do it in headless mode

Cool, dont forget to install playwright

malto101:
Also, related to test plan,

i am not asking for you to implement test cases(though that would be nice :) ), but instead create test scenarios from the feature list,

currently, we just keep adding features but we have no idea of all the features optics provides, this will cause problems in future when we try adding new feature or checking if our framework can handle the set problem statement.

malto101:
@kun-codes

I would suggest changing your pull setting to rebase

In this repo, we use rebase strats and not merge strats

PR #352, opened 2026-07-14
feat/add playwright to optics setup

kun-codes:
- **feat(helper): add Playwright support with Chromium installation**
- **feat(helper): show playwright install command's output in `optics setup` command**

### DIFF
--- a/optics_framework/helper/setup.py
+++ b/optics_framework/helper/setup.py
@@ -139,11 +139,11 @@ def install_packages(requirements: List[str]) -> None:
             print("Installing Playwright Chromium driver along with system dependencies...")
             subprocess.run(  # nosec B603
                 [sys.executable, "-m", "playwright", "install", "--with-deps", "chromium"],
-                capture_output=True, text=True, check=True, shell=False)
+                check=True, shell=False)
 
         print("Drivers installed successfully!")
     except subprocess.CalledProcessError as e:
-        print(f"Installation failed: {e.stderr}")
+        print(f"Installation failed: {e.stderr or e}")
     finally:
         if os.path.exists(req_file):
             os.remove(req_file)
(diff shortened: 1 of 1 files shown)


optics_framework/helper/setup.py
```diff
@@ -130,17 +131,22 @@ def install_packages(requirements: List[str]) -> None:
...
-            [sys.executable, "-m","pip", "install", "-r", req_file], capture_output=True, text=True, check=True, shell=False)  # nosec B603
-        if result.returncode == 0:
-            print("Drivers installed successfully!")
-        else:
-            print("Installation failed!")
+        subprocess.run(  # nosec B603
+            [sys.executable, "-m", "pip", "install", "-r", req_file],
+            capture_output=True, text=True, check=True, shell=False)
 
+        if "playwright" in requirements:
```
thakur-patel, line 138: why is this done separately here? where/how are the other dependencies like appium easyocr installed?

kun-codes: I had to make a special case for playwright since as per playwright [docs](https://playwright.dev/python/docs/browsers#install-system-dependencies) a separate command has to be run after installing it via pip

thakur-patel: add this as developer comments since it is a special case

kun-codes: developer comments as in comment the code?

thakur-patel: yes, exactly that

kun-codes: done
check commit https://github.com/mozarkai/optics-framework/pull/352/commits/796705e2a641519446a1ec43d1f3d6f529578688

thakur-patel (approved):

PR #353, opened 2026-07-14
feat/run tests on pr 2

kun-codes:
- **ci: run tests on pull request**
- **ci: update `optics setup` step to include Playwright, EasyOCR and Appium installation**

Refer to #350 
I had to create a new pr because that branch was outdated compared to main by more than 40 commits. Creating this new pr was easier than updating #350's feature branch

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,5 +26,8 @@ jobs:
       - name: Install dependencies
         run: poetry install --no-interaction --with dev,test
 
+      - name: Run `optics setup`
+        run: poetry run optics setup --install "EasyOCR" "Appium" "Playwright"
+
       - name: Run tests
         run: poetry run pytest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,10 +13,10 @@ jobs:
       contents: read
     steps:
       - name: Check out repository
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Install dependencies
+        run: poetry install --no-interaction --with dev,test
+
+      - name: Run `optics setup`
+        run: poetry run optics setup --install "EasyOCR" "Appium" "Playwright"
+
+      # have to use `xvfb-run` because github actions doesn't have an XServer
+      # tests/feature/engine/test_playwright_youtube.py uses the file from
+      # optics_framework/samples/playwright/config.yaml for the config which uses headed mode
+      # `xvfb-run` fixes this without having to change the config
```
malto101, line 35: that's a bad code insert,
lets have a issue on this,

"a config.yml from samples is being referred in tests "

malto101: #356

chinmayajha:
tests already run on every PR via .github/workflows/tests.yml on main, so the core ask here is covered. also this version references tests/feature/engine/test_playwright_youtube.py which doesn't exist on main. closing, happy to add the optics setup step to the current workflow if we want it.

PR #362, opened 2026-07-15
Supervisor for scaling

thakur-patel:


malto101:
Hello @thakur-patel 
lets not intertwine optics with this, keep all within tool/, 
lets not move these main docker file and helper as we know its not the best effort change.

as its a tool, lets just keep it as tool and its own docker file within tools/supervisor/ folder
also move tests under test folder,

other then the sonarqube warnings, I dont see any other major problems with the approach

malto101:
you are missing https dep

malto101:
I dont see these yamls that you specified in docs

malto101:
pin version of kubectl

malto101:
dont use Open, as its synchronous but apis ares async

malto101:
A new httpx.AsyncClient() is created and torn down for every forwarded request instead of a shared/pooled client. Works, but adds avoidable per request overhead for a scaling-focused front tier.

malto101:
Pool-mode workers are always registered/started at http://127.0.0.1:<port>. Combined with --store redis for multi-replica scaling (the doc's own first example, supervisor_usage.md:15), each replica writes its own-host-local loopback address into the shared store. Any other replica that picks that route will try to reach its own loopback, hitting either nothing or an unrelated local worker on that port

pyproject.toml
```diff
@@ -49,10 +49,12 @@ google-genai = { version = ">=2.0.0,<3.0.0", optional = true }
 # >=3.4.2: earlier 3.4.x omit the starlette>=1.0 pin, letting the resolver keep
 # starlette 0.48 (a combo fastmcp's own code is incompatible with at runtime).
 fastmcp = { version = ">=3.4.2,<4.0.0", optional = true }
+redis = { version = ">=5.0,<7.0", optional = true }
```
kun-codes, line 52: latest version of redis is 8.0.1 

we should make version `^8.0.0`
https://pypi.org/project/redis/

pyproject.toml
```diff
@@ -68,6 +70,7 @@ pylint = ">=3.3.7,<5.0.0"
 pytest = ">=8.3.4,<9.1.0"
 pytest-cov = ">=6.0,<7.1"
 pytest-mock = ">=3.14,<3.16"
+fakeredis = "^2.23"
```
kun-codes, line 73: latest version is `2.36.2`. I feel using the latest version when adding a new dependency is alright

kun-codes (changes requested):

PR #376, opened 2026-07-21
fix(serve): wire AI self-heal through optics serve and optics mcp

### LINKED ISSUE #375: AI self-heal cannot be enabled via optics serve / optics mcp
## Problem

`optics execute` and `optics live` both support AI self-heal via `ai_self_heal: true` + an enabled `llm_models` entry in a project's `config.yaml`. `optics serve` and `optics mcp` (which wraps the same session-creation code) do not — not because it's unreported, but because it's unreachable:

- `SessionConfig`, the request body both `POST /v1/sessions/start` and the MCP `start_session` tool use, has no `ai_self_heal` or `llm_models` fields at all — only `driver_sources`, `elements_sources`, `text_detection`, `image_detection`.
- `create_session` builds a fresh `Config(...)` from exactly those four fields. Even passing `project_path` at a project whose own `config.yaml` already has `ai_self_heal: true` doesn't help — that file is never read for engine-level settings; `project_path` is only used for template discovery and the output-directory default.

Net effect: there is currently no way to turn self-heal on for a session created via `optics serve` or `optics mcp`, regardless of project configuration.

## Where

- `optics_framework/common/expose_api.py` — `SessionConfig`, `create_session`
- `optics_framework/helper/mcp_server.py` — `start_session` (wraps the same `create_session`)

## Expected

A session started via `optics serve`/`optics mcp` should be able to opt into self-heal, either explicitly in the request or by inheriting it from a project's `config.yaml` when `project_path` is given — matching how `execute`/`live` already behave on the same project.

chinmayajha:
Closes #375. Depends on #303 .

## What was missing

`SessionConfig` (used by both `POST /v1/sessions/start` and the MCP `start_session` tool) had no `ai_self_heal`/`llm_models` fields, so a session created that way could never turn self-heal on, even with `project_path` pointing at a project whose own `config.yaml` already had it configured.

Separately, `mode=keyword` execution (the path `serve`/`mcp` use) goes through `KeywordExecutor` — a third call site distinct from `TestRunner`/`LiveController` — which had no visibility into whether a keyword only passed because self-heal recovered it.

## Changes

- `SessionConfig` gains `ai_self_heal`/`llm_models`. `_resolve_self_heal_settings` resolves each independently: an explicit value in the request always wins; otherwise, when `project_path` is given, it's read straight from that project's `config.yaml` — so a project already set up for self-heal behaves the same via `serve` as via `execute`/`live`.
- `KeywordExecutor.execute` now pops the heal info recorded on the `ActionKeyword` instance right after a successful call, publishing it on the keyword `Event`'s `extra` and folding it into the returned value as `{"result": ..., "healed": true, "heal_summary": ...}` when a heal occurred (unhealed calls are unaffected).
- MCP's `start_session` gets a single `ai_self_heal: Optional[bool] = None` param (left unset, it inherits `project_path`'s config; the LLM provider itself still comes from that config or the environment, not this tool — it's a nested provider/credentials shape that doesn't fit a flat tool param). `_make_keyword_tool`'s result-unwrapping is fixed so it no longer silently drops `healed`/`heal_summary` before they reach the MCP client.

### DIFF
--- a/optics_framework/api/action_keyword.py
+++ b/optics_framework/api/action_keyword.py
@@ -4,7 +4,7 @@ import inspect
 import shlex
 import time
 import json
-from typing import Callable, Optional, Any, Tuple, List
+from typing import Callable, Optional, Any, Tuple, List, Dict
 from optics_framework.common.logging_config import internal_logger
 from optics_framework.common.optics_builder import OpticsBuilder
 from optics_framework.common.strategies import StrategyManager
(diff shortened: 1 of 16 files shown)


chinmayajha:
Closing — the enablement design here (per-request project_path pointing at a server-side config.yaml) is the wrong shape for how this API is actually consumed: a platform integrating over HTTP has no business handing this server a filesystem path to read, which is exactly what CodeQL flagged. Redesigning as a service-level (env-var driven) setting instead, in a follow-up PR against #375.

Issue #367, opened 2026-07-20
feat: add a keyword called `assert_visibility`


kun-codes:
**Is your feature request related to a problem? Please describe.**
We have a similar keyword called `assert_presence` but it returns a boolean value depending on if the passed element is present anywhere in the page source. So an element which is not present on the display screen but present out of the viewport is also said to be present. 

`assert_visibility` will return boolean depending on if the passed element is present within the user's display screen

PR #378, opened 2026-07-21
fix(engines): remove fuzzy XPath matching fallback in `AppiumPageSour…

kun-codes:
…ce` to stop false positives on non-existent elements

### DIFF
--- a/optics_framework/engines/drivers/appium_UI_helper.py
+++ b/optics_framework/engines/drivers/appium_UI_helper.py
@@ -79,7 +79,13 @@ class UIHelper:
 
     def find_xpath(self, xpath):
         """
-        Process the given XPath and return the exact path from the UI tree after applying various matching strategies.
+        Process the given XPath and return the exact path from the UI tree, requiring
+        a genuine match (exact, or a simplified/relative form of the same predicate).
+
+        Deliberately does not fall back to partial/fuzzy attribute matching: those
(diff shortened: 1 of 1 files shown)


kun-codes:
@malto101  What do you think about this PR? This pr will remove fuzzy matching on XPath

The reason I did it because when some elements have very similar characteristics, referring to it via XPath can lead to false positives of matching our passed element with something else. This happened to me when implementing `assert_visibility` keyword for #367. When trying to check if "Option 47" is visible on the screen via xpath, the fuzzy matching of XPath falsely concluded that "Option 47" is visible on the screen as the XPath of "Option 1" is very similar to that of "Option 47" as most of the string would be the same but the number differs. I have attached the screenshot of the example below

Since XPath is something programmatic and not fetched via OCR which can be wrong at times, I feel XPath shouldn't have fuzzy matching at all

image

malto101:
Hello @kun-codes 
I see the problem, seems like Levenshtein distance was quite small and gave the false positive.

This seems more like a logic problem 
How about doing a exact match then move fuzzy match, maybe update it to tokenize so we can do token fuzzy matching


The reason I don't want to remove fuzzy is because many a time, we don't provide the full element and as the framework is low code , having it help in case of typos gets treated as a feature/resilience.


Other ideas are welcome

thakur-patel:
@kun-codes let's close the PR and create an issue on this

PR #379, opened 2026-07-21
fix(serve): wire AI self-heal through optics serve and optics mcp as a service-level setting

### LINKED ISSUE #375: AI self-heal cannot be enabled via optics serve / optics mcp

chinmayajha:
Closes #375.
Depends on #303 

## Design

Self-heal is now a **service-level** setting for a running `optics serve`/`optics mcp` process, configured once via environment variables when it starts — not a per-request field requiring the platform to change any of its existing integration code:

- `OPTICS_AI_SELF_HEAL` — enables it (`true`/`1`/`yes`/`on`, case-insensitive)
- `OPTICS_LLM_PROVIDER` — which `llm_models` entry to use (default `gemini`, the only implemented provider today)
- `OPTICS_LLM_MODEL` — optional model override

The operator running the service for a platform sets these once; every session that platform creates inherits self-heal automatically. `SessionConfig.ai_self_heal: Optional[bool] = None` remains as a per-session override (explicit request value always wins over the service default) — there's no `llm_models` field on the request body at all, since provider/model selection is a deployment decision, not a per-request one. LLM credentials continue to come entirely from the provider's own environment variables (e.g. `GeminiLLM` already reads `GOOGLE_API_KEY`/`GEMINI_API_KEY` itself) and never flow through this config.

This removes the CodeQL-flagged code rather than patching it — there's no longer any server-side file read driven by request data.

## Changes

- `optics_framework/common/ai_self_heal.py` / `optics_framework/api/action_keyword.py` — `HealResult.steps_taken` + `ActionKeyword._pop_last_heal_info()`, unchanged from #376 (pure reporting plumbing, independent of enablement).
- `optics_framework/common/execution.py` — `KeywordExecutor.execute` surfaces a heal via the Event and the returned result, unchanged from #376.
- `optics_framework/common/expose_api.py` — `_service_self_heal_settings()` replaces the old project_path-based resolution entirely.
- `optics_framework/helper/mcp_server.py` — `start_session`'s `ai_self_heal` param and the `_make_keyword_tool` result-unwrap fix are unchanged; only the docstring now describes the env vars.
- `docs/usage/mcp_usage.md` — documents the three env vars and the per-session override.
- `tests/units/common/test_expose_api_self_heal.py` — new: env-var resolution (default off, truthy/falsy variants, custom provider/model) and explicit-override-wins-over-service-default semantics.

### DIFF
--- a/optics_framework/common/expose_api.py
+++ b/optics_framework/common/expose_api.py
@@ -157,6 +157,12 @@ class SessionResponse(BaseModel):
 class ExecutionResponse(BaseModel):
     """
     Response model for execution results.
+
+    ``data`` normally carries ``{"result": <value>}``. When AI self-heal recovered the
+    keyword, it instead carries ``{"result": <value>, "healed": true, "heal_summary":
+    <str>, "suggested_steps": [{"keyword": <str>, "params": [<str>, ...]}, ...]}``.
+    ``suggested_steps`` is the clean, replayable recovery sequence — the steps a caller
+    can persist to reproduce the recovery without relying on self-heal next time.
     """
 
     execution_id: str
@@ -248,11 +254,19 @@ class SessionConfig(BaseModel):
 
     Use `normalize_sources()` to convert entries into a consistent list of
     {name: DependencyConfig} mappings used by the server internals.
+
+    ``ai_self_heal`` / ``llm_provider`` / ``llm_model`` are optional per-session
+    self-heal overrides, kept as flat scalars (not a nested llm_models block) so no LLM
+    credentials ride the request body. See ``_resolve_self_heal_settings`` for their
+    per-field precedence over the service-level env-var defaults.
     """
     driver_sources: List[Union[str, Dict[str, Any]]] = []
     elements_sources: List[Union[str, Dict[str, Any]]] = []
     text_detection: List[Union[str, Dict[str, Any]]] = []
     image_detection: List[Union[str, Dict[str, Any]]] = []
+    ai_self_heal: Optional[bool] = None
+    llm_provider: Optional[str] = None
+    llm_model: Optional[str] = None
     project_path: Optional[str] = None
     appium_url: Optional[str] = None
     appium_config: Optional[Dict[str, Any]] = None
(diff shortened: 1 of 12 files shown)


malto101:
not sure if keeping it as a env var to be used at the beginning will be of optimal use,
as MCP and serve is expected to cater to multiple people at once, not everyone will like to affix to a model set by the server admin,
it would be better to keep it as a session config and be set at the "create session".

chinmayajha:
> not sure if keeping it as a env var to be used at the beginning will be of optimal use, as MCP and serve is expected to cater to multiple people at once, not everyone will like to affix to a model set by the server admin, it would be better to keep it as a session config and be set at the "create session".

Addressed this in the latest commit. Keeping env as default but now letting per-session request fields override them.

chinmayajha:
> > > not sure if keeping it as a env var to be used at the beginning will be of optimal use, as MCP and serve is expected to cater to multiple people at once, not everyone will like to affix to a model set by the server admin, it would be better to keep it as a session config and be set at the "create session".
> > 
> > 
> > Addressed this in the latest commit. Keeping env as default but now letting per-session request fields override them.
> 
> Any reason we aren't doing per session basis?

It is per session basis only after the latest commit. env vars are optional. ai_self_heal, llm_provider, and llm_model are all fields on SessionConfig (and start_session). The only thing intentionally kept server-side is credentials @malto101

chinmayajha:
@malto101 @thakur-patel  Please let me know if any further changes are required here

optics_framework/common/expose_api.py
```diff
@@ -253,6 +259,19 @@ class SessionConfig(BaseModel):
     elements_sources: List[Union[str, Dict[str, Any]]] = []
     text_detection: List[Union[str, Dict[str, Any]]] = []
     image_detection: List[Union[str, Dict[str, Any]]] = []
+    # AI self-heal settings (resolved by _resolve_self_heal_settings). A single
```
thakur-patel: can we keep this not inside the doc string above

chinmayajha: Fixed. Also in other places.

thakur-patel (approved):

PR #380, opened 2026-07-22
fix(engines): defer `SeleniumDriver` type import to break circular im…

kun-codes:
…port in `selenium_UI_helper.py`


malto101:
just thinking which would be better,
this diff or just not importing SeleniumDriver?

thakur-patel (commented):
LGTM

malto101 (approved):
LGTM

PR #366, opened 2026-07-17
feat/scroll to find options in select dropdown option

kun-codes:
- **chore(dependencies): bump `pytest` to `^9.0.0` across all dependency groups**
- **chore(tests): add `pytest.toml` configuration**
- **chore(tests): remove redundant pytest config from `pytest.ini` and `pyproject.toml`**
- **docs: fix outdated references to pytest configuration files in documentation**
- **bump: version bump to 1.8.7-beta.3**
- **feat(optics)!: add timeout parameter to `select_dropdown_option` function**
- **feat(api): scroll to find options in `select_dropdown_option`**

PR #381, opened 2026-07-22
Feat/assert visibility keyword

### LINKED ISSUE #367: feat: add a keyword called `assert_visibility`

kun-codes:
Fixes #367


kun-codes:
Closing in favor of a same-repo PR from mozarkai/optics-framework:feat/assert-visibility-keyword directly, so it and #366 (stacked on top of it) both reference the same branch instead of two diverging copies (fork vs. upstream).

PR #387, opened 2026-07-23
Stateless optics serve: session layer, Redis store, lease hardening

### LINKED ISSUE #351: Stateless API Layer
# Stateless API Layer (`optics serve`)

**Type:** Feature / Design
**Tracks:** [Help Wanted §1 — Stateless API Layer](https://mozarkai.github.io/optics-framework/contribution/help_wanted/#1-stateless-api-layer)
**Full design:** `docs/contribution/stateless_api_design.md`

## Problem

Today `optics serve` is a single process that holds sessions in an in-memory dict (`SessionManager.sessions`). This pins every session to the process that created it. The moment a second worker or pod is added to gain parallelism, a request load-balanced to a worker that didn't create the session hits `get_session() → None → "Session not found"`.

That single wall blocks horizontal scaling, session recovery, load balancing, and zero-downtime deployments.

## Framing: three layers, one enabler

"Autoscaling `optics serve`" is really three independent concerns — this issue delivers **Layer 1** and designs its seams so Layers 2 and 3 drop in with no call-site rework or schema migration.

| Layer | Concern | Delivers |
|-------|---------|----------|
| **1 (this issue)** | Stateless session layer | Sessions no longer pinned to one process; any instance can reconstruct a session from shared state. |
| **2** | Multi-worker / multi-pod deployment | Run N processes/pods behind a load balancer; each holds no irreplaceable state. |
| **3** | Device/session scheduling | Place sessions on free devices from any instance; scale up to the device/driver fleet. |

**Non-goal:** in-process concurrency (the synchronous driver call chain blocks the event loop, so simultaneous requests serialize) is a *separate* axis, sidestepped by running multiple pods (Layer 2). Called out only so it is not mistaken for a statelessness bug.

## The anticipation principle

> The shared store — keyed by `session_id` — holds a fully serializable `SessionState` describing how to **reconstruct** a session. Every instance treats live `Session` objects as a disposable local cache over that state, and rebuilds them on demand.

Litmus test: **migration and autoscaling are the same code path.** A request landing on an instance without the live runtime *rehydrates* it from the store. `/export`, `/import`, and `/migrate` are thin wrappers over machinery a load balancer exercises implicitly. If migration is a *special* path, the design is wrong.

## Key design decisions

1. **Reattachability is a driver-declared capability, not a design assumption.** Optics isolates every backend behind `DriverInterface`. Whether a live session can be resumed on another instance is answered by each driver via one contract:
   ```python
   supports_session_migration: bool           # default False — existing drivers stay sticky until they opt in
   def get_reattach_params(self) -> dict: ...  # driver-defined reattach handle + endpoint + device
   def detach(self) -> None: ...               # drop client handle without ending backend session; default = terminate()
   ```
   The framework never inspects the backend or hard-codes which products are reattachable. A driver whose backend accepts a fresh client reattaching declares `True` (migrates freely); everything else (in-process state, or a backend that refuses reattach — e.g. cloud device farms) stays **sticky** and pinned to its origin instance. Sticky sessions can still be **re-created** from the serializable recipe; the capability gate only governs whether the *live* session moves.

2. **Split `Session` into state + runtime.** `SessionState` (durable, serializable Pydantic) vs `SessionRuntime` (non-serializable handles: `OpticsBuilder`, live driver, event SDK/queue, temp dirs). This split is what makes all three layers coherent.

3. **`SessionStore` interface with lease methods defined now.** Layer 1 ships only `InMemorySessionStore` (leases always granted), but call sites already `acquire`/`renew`/`release`. Layer 2 adds `RedisSessionStore` as a pure new class + config switch — zero call-site churn.

4. **`SessionManager` = cache + rehydrator.** Every endpoint calls `get_or_rehydrate(sid)` instead of `sessions.get()` — that one substitution is what makes the API layer stateless.

5. **Strict attach.** Reconstruction must use a fail-closed attach mode so a failed reattach **raises** (never silently launches a fresh session on a possibly-wrong target). "Strict attach" is a mode the reattach contract exposes.

6. **`/migrate` = `detach`.** Once store + rehydrate + detach exist, in-cluster migration is *emergent*: detach releases the lease and leaves state + backend session alive; the next request rehydrates anywhere. `/migrate` refuses non-migratable sessions.

## Scope for this issue (Layer 1)

**In:** state/runtime split, `SessionStore` interface + `InMemorySessionStore`, `get_or_rehydrate`, driver capability gate with safe defaults, `detach` vs `terminate`, strict attach, `resolve_driver_binding` seam, `/export` · `/import` · `/migrate` endpoints, migrating process-local globals (`workspace_hashes`) into `SessionState`.

**Out (hooks left in place):** Redis store, device scheduler, active-execution handoff. Idle-only migration for now — refuse `detach`/`migrate` while a session is `busy` or a stream is open.

## Suggested PR sequence

- [ ] **1. State/runtime split + `SessionStore` interface + `InMemorySessionStore`** — pure refactor, no behavior change. De-risks everything.
- [ ] **2. `export_state` / `create_session_from_state` + `/export` + `/import`** — same-instance round-trip test (export, detach without ending session, import, keep driving).
- [ ] **3. `detach` + driver capability gate + strict attach + instance-local cleanup.**
- [ ] **4. `resolve_driver_binding` seam + fully-populated `DriverBinding`** — the Layer-3 anchor, cheap to add now.

## File-by-file change map

- `common/models.py` — `SessionState`, `DriverBinding`, `SessionStatus`.
- `common/session_manager.py` — `SessionStore` (ABC + `InMemorySessionStore`); refactor to store + cache + rehydrate; `detach`/`terminate`; `instance_id`; `resolve_driver_binding`; `export_state`/`create_session_from_state`.
- `common/optics_builder.py` — accept reattach params / `DriverBinding`; thread `strict` into the attach path.
- `common/driver_interface.py` — the migration contract with safe defaults (drivers stay non-migratable until they opt in). Normative behavior lives here.
- `engines/drivers/<driver>.py` — per driver, *optionally* implement the contract to opt in. Pick one reattachable driver as the reference implementation first.
- `common/expose_api.py` — `/export`, `/import`, `/migrate`; construct `session_manager` with an env-selected store; migrate `workspace_hashes` into state.

## Open decision

**Inline template bytes vs reference in `/export`.** Recommendation: inline base64 in Layer 1 with a documented size cap; add reference-based export as a Layer-2 option once shared object storage exists.

---

*Full design with the seams table, reconstruction detail, and the capability-gate rationale: `docs/contribution/stateless_api_design.md`.*

thakur-patel:
## Summary

Makes `optics serve` stateless so it can run behind a load balancer across
multiple workers/pods, and hardens the shared-store lease semantics for
production. Sessions are no longer pinned to the process that created them:
durable `SessionState` in a shared store is the truth, and any instance
rebuilds (rehydrates + reattaches) a live session on demand. Migration and
autoscaling are the *same* code path.

Design: `docs/contribution/stateless_api_design.md`.

## What's included

**Layer 1 — stateless session layer** (`ed9aa5b`)
- Split `Session` into serializable `SessionState` + disposable `SessionRuntime`; `SessionStore` ABC + `InMemorySessionStore`; `SessionManager` becomes cache + rehydrator (`get_or_rehydrate`).
- Driver capability gate on `DriverInterface`: `supports_session_migration`, `get_reattach_params()`, `reattach(strict=…)`, `detach()` (default = `terminate()`). Appium implements it as the reference migratable driver.
- `/export`, `/import`, `/migrate` endpoints — `/migrate` collapses to `detach`; in-cluster migration is emergent, not a bespoke path.

**Layer 2 — multi-worker/pod shared store** (`8632c73`, `c78216c`)
- `RedisSessionStore` behind the same `SessionStore` seam (zero call-site churn), selected via `OPTICS_SESSION_STORE=redis` / `OPTICS_REDIS_URL`. Optional `stateless` extra.
- Enforced distributed leases; cross-instance conflict → `SessionOwnedElsewhere` → **HTTP 409**. Orphan reclaim is automatic via lease-key TTL.

**Layer 2 hardening** (`0ff4242`, `348c4d3`, `99c7879`)
- **Atomic lease CAS** — `renew`/`release`/acquire-extend are single Lua scripts (compare-owner-then-mutate), closing the check-then-act window where a GET-then-SET/DEL could steal or delete a lease reassigned between the two calls. Fresh acquire stays `SET NX PX`.
- **Keepalive heartbeat** — `SessionManager` renews all live local sessions' leases every ~`TTL/3` from the FastAPI lifespan, so a live-but-idle session isn't reclaimed for lack of traffic. No-op under the in-memory store. TTL configurable via `OPTICS_LEASE_TTL_S`.
- **Store-outage handling** — bounded `redis-py` `Retry` in `from_url`; exhausted → `SessionStoreUnavailable` → **HTTP 503** (retryable), without the API layer importing `redis`.

## Config / ops surface

| Env var | Default | Meaning |
|---|---|---|
| `OPTICS_SESSION_STORE` | `memory` | `memory` \| `redis` |
| `OPTICS_REDIS_URL` | `redis://localhost:6379/0` | Redis URL (redis backend only) |
| `OPTICS_LEASE_TTL_S` | `300` | Lease TTL seconds |

`pip install optics-framework[stateless]` adds `redis`. Default (in-memory) path never imports it.

## Testing

- `tests/units/common/test_stateless_sessions.py`: 35 tests — state/runtime split, capability gate, detach vs terminate, export/import, strict-reattach rehydrate, and the Redis store (CRUD, lease mutual-exclusion, orphan reclaim, CAS ownership contract, keepalive TTL extension, outage→503). Redis path runs against `fakeredis[lua]` (in-process, incl. Lua) — no live Redis needed.
- Full `tests/units/common` + `tests/units/api`: 198 passed, no regressions. Pre-commit clean on every commit.

## Gap analysis / known limitations & follow-ups

Resolved in this branch (were open after initial Layer 2): benign→real lease
race (now atomic Lua CAS), Redis reconnection (bounded retry→503), `poetry.lock`
relock for the extra.

Open, deliberately deferred:

1. **CAS atomicity is not unit-provable.** The tests validate the sequential *ownership contract*; the concurrent-race benefit of Lua can't be reproduced single-threaded. Only an integration test against a live Redis with real concurrency exercises it. *(Follow-up: integration test in CI with a Redis service container.)*
2. **Heartbeat runs sync store calls on the event loop.** Renewals are quick and single-process concurrency is a stated non-goal, but a large per-pod session count could add small periodic loop stalls. *(Follow-up: `run_in_executor` if it surfaces.)*
3. **Heartbeat lease-loss only logs, doesn't evict.** A session whose lease is lost mid-life (e.g. Redis down > TTL) is detected on the next request (409), not proactively evicted — matching the existing detect-on-lookup behavior; a brief two-place window can exist until then.
4. **`redis` widened to `<9.0`; real redis 8.x untested here.** The store's `Retry`/`from_url` path is validated only structurally (tests use fakeredis). *(Follow-up: smoke test against real redis 8.)*
5. **Graceful shutdown doesn't release leases.** `stop_heartbeat` cancels the loop; a terminating pod's leases rely on TTL expiry for reclaim. Faster rolling-deploy failover would need a shutdown hook that detaches live sessions — dovetails with SIGTERM handling for the planned orchestration service.
6. **Lease loss mid-SSE-stream** degrades to a stream error rather than a clean 409 (the handler only covers request-level raises) — a Layer-1 scope boundary (streams are instance-local; migration is idle-only).
7. **Event-stream / active-execution handoff across pods** is out of scope (Layer-1 non-goal); migration is idle-only.

## Out of scope (deferred)

- **Layer 3 (device fleet)** — a pluggable, farm-agnostic `DeviceProvider` seam with the Mozark Appium Hub as first adapter — is designed (`docs/contribution/mozark_appium_hub_integration.md`) but not implemented. Deferred until a second hub or a hub-agnostic client contract is needed; the `resolve_driver_binding` seam is already in place for it.

Closes #351.


optics_framework/common/expose_api.py
```diff
@@ -1143,104 +1214,108 @@ async def workspace_generator(
...
-                    execution_id=EXECUTION_ID_HEARTBEAT,
-                    status=STATUS_HEARTBEAT,
-                    message="No new event, sending heartbeat"
+                    execution_id=EXECUTION_ID_UNKNOWN,
+                    status=STATUS_ERROR,
+                    message=f"AttributeError: {attr_err}"
                 ).model_dump())}
-            except Exception as exc:
-                internal_logger.error(f"Unexpected error while waiting for event: {exc}")
+                break
```
malto101, line 1302: doesnt Attribute error come under exception?

optics_framework/common/session_manager.py
```diff
@@ -1,19 +1,36 @@
...
+    TestCaseNode,
+    ElementData,
+    ApiData,
+    ModuleData,
+    TemplateData,
+    ErrorDefinitions,
+    DriverBinding,
+    SessionState,
+    SessionStatus,
+)
```
malto101, line 26: we better fix to 1 style in future,
have seen this change too many time across multiple people,
high time we tell which linter/style to use

optics_framework/common/session_manager.py
```diff
@@ -51,6 +68,153 @@ def _maybe_setup_junit(
     setup_junit(session_id, config)
 
 
+def resolve_driver_binding(config: Config) -> DriverBinding:
+    """
+    The single seam through which driver-endpoint selection flows
+    (stateless design §7). Layer 1 reads the first enabled driver source
```
malto101, line 74: would not recommend to ref in this way for doctoring

optics_framework/common/session_manager.py
```diff
@@ -118,7 +303,7 @@ def __init__(self, session_id: str, config: Config,
         self.error_definitions = error_definitions
         self.request_template_overrides: Dict[str, str] = {}
         self.inline_templates: Dict[str, str] = {}
-        self._inline_templates_dir: str = tempfile.mkdtemp(prefix="optics_session_")
+        runtime.inline_templates_dir = tempfile.mkdtemp(prefix="optics_session_")
```
malto101, line 306: type it to str

optics_framework/common/session_store_redis.py
```diff
@@ -0,0 +1,191 @@
+"""Redis-backed :class:`SessionStore` — Layer 2 of the stateless API design.
```
malto101, line 1: as I expect other stores to come,
move redisstore and "in memory store" to a folder called session store under common folder

optics_framework/common/session_manager.py
```diff
@@ -160,25 +413,365 @@ def create_session(self, config: Config,
...
+                raise SessionOwnedElsewhere(session_id)
+            return session
+        state = self.store.get_state(session_id)
+        if state is None or state.status == SessionStatus.TERMINATED:
+            return None
+        if not self.store.acquire_lease(session_id, self.instance_id, self.lease_ttl_s):
+            raise SessionOwnedElsewhere(session_id)
+        session = self._reconstruct_runtime(state)
+        self.sessions[session_id] = session
+        return session
```
malto101, line 452: failed rehydration leaves the lease held

I would suggest to make it something like
```
if not self.store.acquire_lease(...):
    raise SessionOwnedElsewhere(session_id)
try:
    session = self._reconstruct_runtime(state)
except Exception:
    self.store.release_lease(session_id, self.instance_id)
    raise
```

optics_framework/common/session_manager.py
```diff
@@ -160,25 +413,365 @@ def create_session(self, config: Config,
...
+        self.sessions.pop(session_id, None)
+        self._teardown_local_runtime(session)
+        cleanup_junit(session_id)
+        get_event_manager_registry().remove_session(session_id)
+        state.status = SessionStatus.DETACHED
+        state.owner_instance_id = None
+        state.busy = False
+        state.updated_at = time.time()
+        self.store.put_state(state)
+        self.store.release_lease(session_id, self.instance_id)
```
malto101, line 589: migrate/detach does not verify current lease ownership

A pod may retain a stale local runtime after its lease has expired and another pod has rehydrated it. A late /migrate request to that old pod can overwrite the new owner’s state as DETACHED.

optics_framework/common/session_manager.py
```diff
@@ -160,25 +413,365 @@ def create_session(self, config: Config,
...
-            if base_dir:
-                try:
-                    shutil.rmtree(base_dir)
-                except OSError as e:
-                    internal_logger.warning("Failed to remove inline templates directory %s: %s", base_dir, e)
+            self._teardown_local_runtime(session)
+        cleanup_junit(session_id)
+        get_event_manager_registry().remove_session(session_id)
+        self.store.release_lease(session_id, self.instance_id)
+        self.store.delete_state(session_id)
```
malto101, line 605: One distributed-ownership concern to address before merging: terminate_session() acts on the local runtime and deletes shared state without confirming that this instance still owns the session lease.
In the normal stateless flow, a pod’s in memory Session is only a disposable cache; Redis lease ownership determines which pod is currently authorized to control the live backend session. If this pod loses Redis connectivity long enough for its lease to expire, another pod may acquire the lease and reattach to the same Appium session. A delayed /stop handled by the original pod could then call driver.terminate() and delete the state for a session now owned by the other pod.
Could we fence destructive lifecycle operations (terminate, and ideally detach/migrate) by renewing/verifying the lease immediately before touching the driver or store? If renewal fails, discard the stale local runtime and return a 409 instead. A Redis-side atomic “verify owner + transition state/release lease” operation would make the detach path robust too.
Please add a regression test covering: pod A loses its lease, pod B acquires and rehydrates, then pod A attempts /stop or /migrate; pod B’s backend session and stored state must remain intact.

malto101 (changes requested):
the layers feels like layers and more like completion stages
but the direction is correct

confirm me few things

PR #388, opened 2026-07-23
test: production-grade overhaul of the test suite

chinmayajha:
## What & why

The test suite had drifted: it grew alongside features but not in an organised way, hurting readability and hiding gaps. This is a full overhaul toward production-grade hygiene, delivered as 10 individually-reviewable commits (green after each).

## Highlights

- **Baseline → now:** 358 → **524 passing** tests, suite **7,740 → 5,837 lines** (leaner *and* broader).
- **`test_generate.py`: 3,410 → 528 lines.** It was 44% of the suite and ~60% repetition (a copy-pasted {CSV,YAML,mixed}×{pytest,robot} matrix, mock-pandas tests asserting only `isinstance(dict)`, and no-op verifiers that literally could not fail). Now parametrized; generate.py coverage **85% → 95%**.
- **Deleted 5 dead/flaky files** — one collected 0 tests (a `__main__` smoke script), one tested a `VisionInterface` class that doesn't exist, one asserted the mock's own side-effect; plus a live-`youtube.com` test and a ~100MB-model-download OCR test. The suite is now fully **hermetic**.
- **Marker discipline made real:** `-m white_box` now selects the whole unit suite (auto-applied by path in a root `conftest.py`), and the mock API server binds an ephemeral port via a session fixture (no more fixed-8001 collisions / thread leaks).
- **First-ever coverage** of the execution core (`test_runnner.py`), the HTTP surface (`expose_api.py` 0→79%), `YAMLDataReader` + error definitions (30→82%), and the real strategy classes (35→53%, incl. `ImageDetectionStrategy`).

## Two real bugs surfaced (verified)

The overhaul found — and the *old* tests were hiding — two shipped bugs:

- #385 — module conditions decide truth by whether the module *raised*, not `bool(result)`.
- #386 — the param-fallback ladder's `str(oe.code).startswith("E02")` check is always False, so `E0201` never advances the ladder.

Both are fixed in stacked PRs based on this branch (each flips an `xfail(strict)` test added here).

## Notes for review

- No behaviour changes to product code in this PR — tests only.
- Two `xfail(strict=True)` tests pin #385/#386 until their fix PRs land.

Issue #386, opened 2026-07-23
Param-axis fallback ladder never advances on E0201 (str(code) vs code.value bug)


chinmayajha:
## Summary

The param-axis fallback ladder ("fallback level 1" in CLAUDE.md) never advances on an `E0201` element-not-found error. The advance predicate at `optics_framework/common/runner/test_runnner.py:451` is:

```python
if str(oe.code).startswith("E02") or oe.code == Code.X0201:
    continue  # advance to the next ${var} candidate
```

`Code` is a `(str, Enum)`, so `str(Code.E0201)` evaluates to `"Code.E0201"` — **not** `"E0201"`. Therefore `.startswith("E02")` is **always False**, and only `X0201` can advance the ladder.

### Empirical confirmation

```
>>> from optics_framework.common.error import Code
>>> str(Code.E0201)
'Code.E0201'
>>> str(Code.E0201).startswith("E02")
False
>>> Code.E0201.value
'E0201'
```

## Impact

`E0201` is the most common element-not-found code — it is what `StrategyManager.locate` raises when the strategy generator yields nothing (`strategies.py:628`). Because it is treated as **fatal** rather than advancing, a keyword that fails to locate an element under the first `${var}` candidate aborts the whole ladder instead of trying the next fallback value. This defeats the documented level-1 fallback for the single most common failure mode.

The same bug is present in `PytestRunner._try_execute_with_fallback_pytest` (around `test_runnner.py:996`).

## Fix

Compare the enum's value, e.g. `oe.code.value.startswith("E02")` (or match against the element-not-found codes explicitly). Fix both call sites.

## Test coverage already in place

`tests/units/common/test_test_runner.py::TestFallbackAdvanceRule`:
- `test_e0201_does_not_advance_current_buggy_behaviour` pins the current behaviour green.
- `test_e0201_should_advance_the_ladder` is `xfail(strict=True)` — it will flip to XPASS (failing the run, prompting removal of the marker) once the predicate is fixed.

## Note on scope

Behaviour change to the execution core — ship in its own PR with a changelog note, not folded into unrelated work.

_Surfaced by the test-suite overhaul while writing the first real coverage of the execution core._

chinmayajha:
Fix is up in #399 (rebased onto `main` after #396 merged; directly mergeable). It compares `oe.code.value` at both call sites and flips the `xfail(strict)` `test_e0201_should_advance_the_ladder` guard to passing. This issue will auto-close when #399 merges.

Issue #385, opened 2026-07-23
FlowControl module condition decides truth by exception-presence, not result truthiness


chinmayajha:
## Summary

`FlowControl._handle_module_condition` (`optics_framework/api/flow_control.py:335`) decides whether a **module-name condition** is true by whether `execute_module(cond)` *raised*, not by the truthiness of its result. A module that runs successfully but returns an empty/falsy result (e.g. `[]`) is therefore treated as **true**, and its target module runs anyway.

## Why it's a bug

This contradicts:

1. **The documented contract** — `_evaluate_conditions` docstring: *"For module-based conditions, returns the result of executing the module if non-empty."*
2. **The sibling expression path** — `_is_condition_true` (`flow_control.py:389`) correctly does `result_bool = bool(result)` for the same module-name case.

So `condition("moduleThatReturnsEmpty", "target")` runs `target` today, when it should fall through to the `else` branch (or return `None`).

## Correct contract

A module used as a condition is **true** iff it executes without raising **and** returns a non-empty (truthy) result; **false** if it returns empty/falsy **or** raises. The non-inverted and inverted (`!`) paths should both branch on `bool(result)`, not on exception-presence.

## Test coverage already in place

`tests/units/api/test_flow_control_all.py::TestModuleCondition` encodes the correct contract. The two cases that expose the bug are marked `@pytest.mark.xfail(strict=True)`:

- `test_falsy_module_skips_target_runs_else`
- `test_falsy_module_no_else_returns_none`

They fail today (xfail) and will turn into XPASS (failing the run, prompting removal of the marker) once this is fixed — so landing the fix and deleting the two markers is the acceptance criterion.

## Note on scope

This is a **behaviour change to a shipped keyword** — a suite that currently relies on the buggy "empty module condition still runs target" semantics would change. Ship it in its own PR with a changelog/deprecation note, not folded into unrelated work.

_Filed as part of the test-suite overhaul (analysis surfaced the latent bug)._

chinmayajha:
Fix is up in #398 (rebased onto `main` after the test-overhaul #396 merged; directly mergeable). It flips the two `xfail(strict)` guards added by the overhaul — `test_falsy_module_skips_target_runs_else` and `test_falsy_module_no_else_returns_none` — to passing. This issue will auto-close when #398 merges.

PR #396, opened 2026-07-24
test: production-grade overhaul of the test suite

chinmayajha:
> Supersedes #388 — reopened from a fork branch (`chinmayajha:test/overhaul-analysis`) per the fork-based contribution flow. Content is identical.

## What & why

The test suite had drifted: it grew alongside features but not in an organised way, hurting readability and hiding gaps. This is a full overhaul toward production-grade hygiene, delivered as individually-reviewable commits (green after each).

## Highlights

- **Baseline → now:** 358 → **524 passing** tests, suite **7,740 → 5,837 lines** (leaner *and* broader).
- **`test_generate.py`: 3,410 → 528 lines.** It was 44% of the suite and ~60% repetition (a copy-pasted {CSV,YAML,mixed}×{pytest,robot} matrix, mock-pandas tests asserting only `isinstance(dict)`, and no-op verifiers that literally could not fail). Now parametrized; generate.py coverage **85% → 95%**.
- **Deleted 5 dead/flaky files** — one collected 0 tests (a `__main__` smoke script), one tested a `VisionInterface` class that doesn't exist, one asserted the mock's own side-effect; plus a live-`youtube.com` test and a ~100MB-model-download OCR test. The suite is now fully **hermetic**.
- **Marker discipline made real:** `-m white_box` now selects the whole unit suite (auto-applied by path in a root `conftest.py`), and the mock API server binds an ephemeral port via a session fixture (no more fixed-8001 collisions / thread leaks).
- **First-ever coverage** of the execution core (`test_runnner.py`), the HTTP surface (`expose_api.py` 0→79%), `YAMLDataReader` + error definitions (30→82%), and the real strategy classes (35→53%, incl. `ImageDetectionStrategy`).
- **Imports hoisted to module top** across the suite (no mid-function imports; the one guarded playwright import stays local behind `importorskip`).

## Two real bugs surfaced (verified)

The overhaul found — and the *old* tests were hiding — two shipped bugs:

- #385 — module conditions decide truth by whether the module *raised*, not `bool(result)`.
- #386 — the param-fallback ladder's `str(oe.code).startswith("E02")` check is always False, so `E0201` never advances the ladder.

Both are fixed in follow-up PRs from this fork (each flips an `xfail(strict)` test added here); they'll be opened against `main` once this lands.

## Notes for review

- No behaviour changes to product code in this PR — tests only.
- Two `xfail(strict=True)` tests pin #385/#386 until their fix PRs land.


thakur-patel:
sqaush the commits

chinmayajha:
> sqaush the commits

I think as this is an overhaul, squashing should not be done. Each of the commit focus on a large block of tests. @thakur-patel

thakur-patel (approved):
this is good for an initial overhaul, it's better than before
@malto101 we can add any suggestions to the tests structure as an issue and take it up separately
@kun-codes will be working on adding code coverage of our tests in a separate PR

malto101:
I am fine with the commits here, 
Just want to move away from the merge commit present here,

Merge commit makes sense when the commits where already pushed to main remote
But here it just creates noise

Rest looks good overall 
Wasn't able to review the code indepth
(Partially due to not having a test plan)

thakur-patel:
> (Partially due to not having a test plan)

@malto101 are you talking about a TESTING.md file for optics-framework?

malto101:
> > (Partially due to not having a test plan)
> 
> @malto101 are you talking about a TESTING.md file for optics-framework?

That's 1 thing
But the main thing is
1. Current tests are created by feels, there is board/maintain them
2. As new features come, tests should be planned against it, but many features go unnoticed/untested
3. There is no place where we can glance what all tests are being covered in the feature list/ APIs list/ end to end usage list

tests/units/engines/ocr_models/test_easyocr_helper.py
```diff
@@ -1,41 +0,0 @@
...
-    "sample_text_image.png"  # Construct absolute path
-    image = cv2.imread(str(image_path))
-    assert image is not None, "Test image not found or failed to load"
-    return image
-
-
-def test_detect_text(easyocr_instance, sample_image):
-    """Test that detect method correctly identifies text in an image."""
-    reference_text = "Connected"
-    result = easyocr_instance.detect_text(sample_image)
```
kun-codes, line 27: **test-coverage (CONFIRMED):** deleting this file (real EasyOCR model, non-hermetic — reasonable) leaves `EasyOCRHelper.detect_text` (`optics_framework/engines/vision_models/ocr_models/easyocr.py:106`) with **zero** test coverage anywhere in the repo. The only remaining `easyocr` references (`tests/units/common/test_expose_api.py`) just check the config-string round-trips, never call `detect_text`. A regression in the returned `(bbox, text, confidence)` shape would ship silently. Worth a hermetic/mocked replacement.

tests/feature/engine/test_playwright_youtube.py
```diff
@@ -1,77 +0,0 @@
...
-        return yaml.safe_load(f)
-
-
-@pytest.fixture(scope="module")
-def optics_instance():
-    """
-    Fixture to setup and teardown Optics with Playwright driver
-    """
-    config = load_config(PLAYWRIGHT_CONFIG_PATH)
-    optics = Optics()
```
kun-codes, line 25: **test-coverage (CONFIRMED):** deleting this (flaky, live-youtube.com, non-hermetic — reasonable) removes the *only* test exercising `engines/drivers/playwright.py` and `engines/elementsources/{playwright_find_element,playwright_page_source}.py`. Verified via `pytest --cov`: 0/237, 0/117, 0/452 lines covered, coverage.py reports each module never imported by any test. (`playwright_screenshot.py` is fine — retains 59% coverage via a pre-existing unrelated test.) A regression in locator resolution, page-source extraction, or press/launch for the Playwright backend now ships with zero signal. No hermetic replacement was added.

tests/units/helpers/test_generate.py
```diff
@@ -1,3410 +1,526 @@
...
-    @pytest.mark.generate
-    @pytest.mark.white_box
-    def test_generate_test_function_special_names(self):
-        """Test test function generation with special characters in names."""
-        result = self.generator._generate_test_function(
-            "Login-Test Case #1", ["Login Module"]
-        )
-
-        assert "def test_login-test_case_#1(optics):" in result
-        assert "login_module(optics)" in result
```
kun-codes, line 972: **test-coverage (CONFIRMED):** this deleted test pinned a real, still-present bug in `PytestGenerator._generate_test_function` (`optics_framework/helper/generate.py:371-379`): it only lowercases + joins on whitespace, no identifier sanitization. Reproduced live: a test case named `"Login-Test Case #1"` still generates `def test_login-test_case_#1(optics):`, which fails `ast.parse` with `SyntaxError`. The lean parametrized rewrite dropped this case with no equivalent — no test in the current suite would catch a CSV/YAML author naming a test case with a hyphen or `#` breaking `optics generate` output.

tests/units/api/test_flow_control_all.py
```diff
@@ -1,405 +1,357 @@
...
-
-def test_read_data_env_json(flow_control, monkeypatch):
-    json_content = json.dumps([
-        {"alpha": "beta", "val": 42},
-        {"alpha": "gamma", "val": 99}
-    ])
-    monkeypatch.setenv('MYJSON', json_content)
-    result = flow_control.read_data('my_elem', 'ENV:MYJSON', "alpha == 'gamma';select=val")
-    assert result == ['99']
-    assert flow_control.session.elements.get_first('my_elem') == '99'
```
kun-codes, line 94: **test-coverage (CONFIRMED):** this deleted test was the only coverage for `FlowControl.read_data`'s env-var-as-JSON-array path (`_load_env_data` → `json.loads` → `_normalize_json_data` → `pd.json_normalize`, `optics_framework/api/flow_control.py:503-511`). The new `TestReadData` class only parametrizes env-scalar/env-CSV cases, none JSON-shaped. `_normalize_json_data` isn't exercised elsewhere either (the file-sourced JSON tests use separate inline normalization logic in `_load_json_file`), so this function/branch is now completely untested and this gap isn't called out in any commit message, unlike the Playwright/EasyOCR removals.

tests/units/api/test_flow_control_all.py
```diff
@@ -1,405 +1,357 @@
...
 
-class DummySession:
+class _Session:
     def __init__(self):
         self.elements = ElementData()
-        self.modules = type('M', (), {'modules': {}, 'get_module_definition': lambda self, x: self.modules.get(x)})()
+        self.modules = _Modules()
         self.apis = ApiData()
         self.apis.collections = {}
+        self.config_handler = MagicMock()
```
kun-codes, line 49: **correctness (PLAUSIBLE):** `_Session.config_handler = MagicMock()` is unconfigured, so `FlowControl._execute_request` (`optics_framework/api/flow_control.py:1071-1077`) stringifies `config_handler.config.execution_output_path` (an auto-created Mock attribute) via `os.fspath`/`os.makedirs`, writing a real `MagicMock/mock.config.execution_output_path/<id>/` directory tree at the repo root on every `TestInvokeApi` run. Reproduced live (21+ leaked dirs from a single verification pass); this PR's own commit b6a1611 added `MagicMock/` to `.gitignore` as a workaround rather than fixing the root cause. Fix: set `config_handler.config.execution_output_path = None` (or a real tmp_path), as `mock_runner` in `tests/units/conftest.py` already does.

tests/units/test_optics.py
```diff
@@ -88,49 +72,43 @@ def test_add_testcase_and_module():
...
+    optics.invoke_api("mock.token")
+    optics.add_element("access_token", optics.get_element_value("access_token")[0])
+    optics.add_element("userId", optics.get_element_value("userId")[0])
+    optics.invoke_api("mock.sendotp")
+    assert optics.get_element_value("txnType")[0] == "GEN"
+    optics.quit()
+
 
-def test_flow_control_methods():
+def test_run_loop_executes_module():
```
kun-codes, line 94: **test-coverage (CONFIRMED):** splitting `test_flow_control_methods` (which called `optics.condition('True', 'mod1')`) into these three tests drops all coverage of the `Optics.condition` SDK-facade wrapper (`optics_framework/optics.py:1258`, incl. its `INVALID_SETUP` guard). Every remaining `.condition(` call in `tests/` now hits `FlowControl.condition` directly, not through the facade — exactly the `optics.py:Optics` vs. API-class drift this repo's CLAUDE.md calls out as a recurring pitfall. Also worth flagging here: each of these 3 tests independently pays `optics.setup()`'s ~2-6s EasyOCR-instantiation cost (contact/config.yaml has `easyocr: enabled: true`) even though none touches OCR — measured 5.80s+2.74s+2.45s ≈ 11s vs. one setup call previously. Consider sharing one setup/quit across these, and adding back a facade-level `condition` call somewhere.

tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+_BOX_MARKERS = {"white_box", "black_box", "hybrid"}
+
+
+def pytest_collection_modifyitems(config, items):
+    """Tag unmarked unit tests ``white_box`` based on their location."""
+    for item in items:
+        path = str(item.fspath).replace("\\", "/")
+        if "/tests/units/" not in path:
+            continue
+        if not any(m.name in _BOX_MARKERS for m in item.iter_markers()):
```
kun-codes, line 27: **conventions (CONFIRMED):** this auto-tagging hook labels every unmarked test under `tests/units/` as `white_box`, but the rewritten `pytest.toml` docstring in this same PR defines `white_box` as "collaborators mocked." Several changed files with no explicit box marker drive real (unmocked) collaborators — `test_data_reader.py` (real file I/O), `test_strategies.py` (commit 18f0c06 explicitly drives real strategy classes), `test_optics.py::test_api_token_and_otp_flow` and `test_api_invocation.py` (both hit the real in-process `mock_api_server`) — and all get auto-tagged `white_box` anyway, while `black_box`/`hybrid` currently select zero tests. Not breaking CI today (no workflow filters by marker), but the marker semantics this PR just wrote are self-contradicted.

tests/units/conftest.py
```diff
@@ -1,150 +1,67 @@
...
 def mock_runner():
-    """Fixture providing a mock runner with temporary directories."""
-    session = MagicMock(spec=Runner)
-    session.elements = ElementData()
-    session.config_handler = MagicMock()
+    """A Runner-spec'd mock carrying real ElementData and a temp output path."""
+    runner = MagicMock(spec=Runner)
+    runner.elements = ElementData()
+    runner.config_handler = MagicMock()
     temp_dir = tempfile.mkdtemp()
```
kun-codes, line 33: **test-hygiene (CONFIRMED):** `mock_runner` calls `tempfile.mkdtemp()` with no cleanup (no `yield`+`shutil.rmtree`, no `tmp_path`), leaking an OS temp dir on every test using this shared fixture. This PR's own commit 75af8f6 fixed the identical leak pattern one file over in `test_live_save.py` ("stop temp-dir leak") — worth applying the same fix here (and to `MockOpticsBuilder.temp_dir` in `tests/units/api/test_action_keyword.py:23`, same issue).

tests/units/common/test_expose_api.py
```diff
@@ -1,91 +1,802 @@
...
+
+
+def test_list_keywords_endpoint(client):
+    resp = client.get("/v1/keywords")
+    assert resp.status_code == 200
+    slugs = {k["keyword_slug"] for k in resp.json()}
+    assert "press_element" in slugs
+
+
+def test_execute_keyword_session_not_found_returns_404(client):
```
kun-codes, line 486: **simplification (CONFIRMED):** this is one of four near-identical "session not found → 404" tests (also at lines 695, 725, 769) — each independently patches `session_manager.get_session` to return `None` and asserts 404 against a different endpoint. All four endpoints raise the identical `HTTPException(404, SESSION_NOT_FOUND)`. Lines 695 and 769 only check `status_code`, skipping the `detail` check the other two make (copy-paste drift). Consider collapsing into one parametrized table over `(method, path, json_body)` so a change to the 404 contract touches one call site instead of four.

PR #398, opened 2026-07-24
fix(api): decide module conditions by result truthiness, not exception-presence

### LINKED ISSUE #385: FlowControl module condition decides truth by exception-presence, not result truthiness

chinmayajha:
Fixes #385.

`FlowControl._handle_module_condition` treated a module-name condition as true whenever the module ran **without raising** — ignoring `bool(result)`. A module returning an empty/falsy result therefore wrongly ran its target (and skipped `else`), contradicting the docstring and the sibling expression path `_is_condition_true`.

This decides truth by `bool(result)` for both the plain and inverted (`!`) forms, and flips the two `xfail(strict)` tests added in the test overhaul to passing (`test_falsy_module_skips_target_runs_else`, `test_falsy_module_no_else_returns_none`).

⚠️ Behaviour change to a shipped keyword: a suite relying on "an empty module condition still runs the target" will change. Full suite green (567 passed).

### DIFF
--- a/optics_framework/api/flow_control.py
+++ b/optics_framework/api/flow_control.py
@@ -333,37 +333,38 @@ class FlowControl:
         return self.modules.get_module_definition(actual_cond) is not None
 
     def _handle_module_condition(self, cond_str: str, target: str) -> Optional[List[Any]]:
-        """Handles evaluation and execution for module-based conditions."""
-        # Strip "!" prefix if present and determine if we need to invert the result
+        """Handles evaluation and execution for module-based conditions.
+
+        A module condition is true iff the condition module executes without
+        raising *and* returns a non-empty (truthy) result — mirroring the
+        expression path in ``_is_condition_true``. A module that returns an
+        empty result is false, so the target is skipped and the caller falls
+        through to the ``else`` branch. The optional ``!`` prefix inverts.
+        """
+        # Strip "!" prefix if present and determine if we need to invert the result.
         invert = cond_str.startswith("!")
         actual_cond = cond_str[1:].strip() if invert else cond_str
 
         try:
-            self.execute_module(actual_cond)
-            # Module executed successfully
-            if invert:
-                # If inverted, success means we should NOT execute target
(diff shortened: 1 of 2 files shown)


malto101:
This is quite a breaking change
Will need some to test this and then come back to this

malto101:
before moving forward with this, 
@chinmayajha could you verify if all existing running tests present in mozark is compatible with this change 
mainly the tests which uses conditions

PR #395, opened 2026-07-24
Overhaul onboarding: extras-based install, fixed init/setup/config, accurate docs

chinmayajha:
## Why
Getting started with Optics was broken end-to-end: a plain `pip install optics-framework` produced a framework whose drivers couldn't import; `optics init` wrote an empty `config.yaml` then a hybrid file layout; `optics setup` clobbered the user's `requirements.txt`; `optics config` silently corrupted dependency lists and edited a file `optics execute` never reads; and the docs described commands and flows that don't exist. This PR makes the first run actually work and rewrites the docs to match.

## What changed (reviewable per-commit)
1. **`fix(samples)`** — remove a committed live API key, drop hardcoded `/Users/...` paths, untrack `.DS_Store`, list all 6 samples in `metadata.json`, make the `playwright` template runnable, and negate the samples subtree in `.gitignore` so sample files stay committable.
2. **`feat(setup)`** — declare every engine backend (appium/selenium/playwright/ble/easyocr/pytesseract/google-vision) as an **optional extra** whose name matches the `config.yaml` source key, plus `mobile`/`web`/`vision`/`all` bundles. `optics setup` now installs `optics-framework[extra]` pinned to the installed version (no CWD `requirements.txt`), accepting both config-key and legacy display names.
3. **`fix(init)`** — write a real commented starter `config.yaml` (was literally `''`), scaffold the sample subdir layout, copy templates verbatim (no more hybrid), skip junk files, and list valid templates on error.
4. **`fix(config,completion)`** — fix the config-TUI bug that keyed every dependency entry as `"name"`; show which (global) file the TUI edits; refresh the shell-completion scripts (drop nonexistent `version`, add `live`/`mcp`/`playwright`, fix the `calender` typo).
5. **`docs`** — new **Installation & Prerequisites** page (Appium server, Node, adb, JDK, browsers, OCR binaries); rewrite README + quickstart + index + CLI/config/workflow docs to match reality (`optics --version`, MkDocs not Sphinx, ruff not Black, correct project layout, accurate global-vs-project config model).
6. **`build`** — regenerate `poetry.lock` for the new extras.

## Verification
- `poetry run pytest`: **524 passed, 1 skipped, 3 xfailed** (skip/xfails all pre-existing).
- pre-commit clean on every commit; `poetry lock` resolves with no conflicts; `mkdocs build` succeeds.
- Manually exercised `optics setup --list`, the driver resolver, both `optics init` paths, and playwright-sample discovery via `find_files`.

## Action required after merge
[...]

PR #401, opened 2026-07-24
fix(samples): remove committed secret, portability paths, and cruft

chinmayajha:
Part **1 of 4** of the onboarding overhaul split out of #395 (per review feedback to break it into reviewable chunks). This chunk is samples hygiene only — it removes a committed secret and makes the samples portable/committable.

## What changed
- **Remove a committed live secret.** Replace the live ipgeolocation API key in the `calendar` sample with a `YOUR_IPGEOLOCATION_API_KEY` placeholder + comment. ⚠️ **The previously committed key must be rotated/revoked at the provider** — it remains in git history.
- **Portability.** Drop the hardcoded `/Users/samyadav` absolute log/project/output paths from the `playwright` sample; paths now default under the project folder.
- **Make the playwright sample runnable.** It previously shipped a config with no tests — add a minimal `test_cases/`, `modules/`, `test_data/elements.csv` suite.
- **Cruft.** Untrack the two committed `.DS_Store` files under `samples/clock`.
- **Discoverability.** List all six samples in `metadata.json` (was only three).
- **`.gitignore`.** Negate the samples subtree so sample project files (csv/yaml/json/jpg) stay committable despite the broad root globs.

## Verification
Full suite green on top of current `main`: **565 passed, 1 skipped, 2 xfailed**.

## Note
Suggested merge order across the split: this (samples) → setup-extras → init/config → docs.


thakur-patel (approved):
LGTM

PR #402, opened 2026-07-24
feat(setup): declare engine backends as extras; install them via optics setup

chinmayajha:
Part **2 of 4** of the onboarding overhaul split out of #395. This chunk fixes the broken install/`optics setup` path by declaring engine backends as optional extras.

## Why
A plain `pip install optics-framework` produced a framework whose drivers/OCR/LLM engines could not import — the third-party packages (`appium-python-client`, `selenium`, `playwright`, `pyserial`, `easyocr`, `pytesseract`, `google-cloud-vision`) were not declared anywhere in `pyproject`. `optics setup` "installed" them by writing a `requirements.txt` into the user's CWD (clobbering any existing one), then deleting it, with unpinned versions and display names that didn't match the config keys.

## What changed
- Declare each engine backend as an **optional dependency** with a version floor + next-major ceiling, mapped to a per-engine **extra whose name matches the `config.yaml` source key** (`appium`/`selenium`/`playwright`/`ble`/`easyocr`/`pytesseract`/`google-vision`), plus `mobile`/`web`/`vision`/`all` convenience bundles.
- Rework `optics setup` to install `optics-framework[<extra>]` **pinned to the installed version** (single source of truth for versions; never upgrades the CLI). No more CWD `requirements.txt`.
- Accept both config-key names (`appium`, `google-vision`) and legacy display names (`Appium`, `Google Vision`) case-insensitively; dedupe aliases.
- `optics setup --list` now prints the install keys.
- Regenerate `poetry.lock` for the new extras (`poetry check --lock` passes).

## Verification
Full suite green on top of current `main`: **565 passed, 1 skipped, 2 xfailed**. `poetry check --lock` clean.

---
### Naming: driver vs engine
`setup.py` now follows the framework's own hierarchy (`optics_framework/engines/`): **engine** is the umbrella for every installable backend, **driver** is the action-driver subtype (`DriverInterface`, `driver_sources`).
- The install-descriptor model is `EngineBackend` (was the misnamed `DriverPackage`→`Driver`); its category container is `EngineCategory`.
- `ACTION_DRIVERS` keeps the "driver" label deliberately (appium/ble/selenium/playwright genuinely are drivers). The OCR and LLM groups are `TEXT_ENGINES` / `LLM_ENGINES` — they aren't drivers.
- Generic helpers/output renamed to the umbrella term: `resolve_engines`, `list_engines`, `EngineInstaller(App)`, "Invalid engine(s)", `optics setup --list` prints "Action Drivers / OCR Engines / LLM Engines".

The sibling completion PR (#403) tracks the `optics setup --list` header rename.

### DIFF
--- a/optics_framework/helper/cli.py
+++ b/optics_framework/helper/cli.py
@@ -9,7 +9,7 @@ from optics_framework.helper.version import VERSION
 from optics_framework.helper.execute import execute_main, dryrun_main
 from optics_framework.helper.live import live_main
 from optics_framework.helper.generate import generate_test_file as generate_framework_code
-from optics_framework.helper.setup  import DriverInstallerApp, list_drivers, install_packages, ALL_DRIVERS
+from optics_framework.helper.setup import EngineInstallerApp, list_engines, install_extras, resolve_engines
 from optics_framework.helper.serve import run_uvicorn_server
 from optics_framework.helper.autocompletion import update_shell_rc
 
@@ -358,33 +358,29 @@ class LiveCommand(Command):
         live_main(live_args.project_folder)
 
 
-class DriverInstaller(Command):
+class EngineInstaller(Command):
     def register(self, subparsers: argparse._SubParsersAction):
         parser = subparsers.add_parser(
-            "setup", help="Install driver for the project")
-        parser.add_argument("--install", nargs="+",
-                            help="Install specified drivers")
+            "setup", help="Install optional engine backends (drivers/OCR/LLM)")
+        parser.add_argument(
+            "--install", nargs="+", metavar="NAME",
+            help="Install the given engines, e.g. --install appium easyocr")
         parser.add_argument("--list", action="store_true",
-                        help="List all available drivers")
+                        help="List all available engines")
         parser.set_defaults(func=self.execute)
 
     def execute(self, args):
         if args.list:
-            list_drivers()
+            list_engines()
         elif args.install:
-            driver_to_install = args.install
-            invalid_drivers = [
-                d for d in driver_to_install if d not in ALL_DRIVERS]
-            if invalid_drivers:
-                print(f"Error: Invalid driver(s): {', '.join(invalid_drivers)}")
-                print("Use --list to see available drivers")
+            engines, invalid = resolve_engines(args.install)
+            if invalid:
+                print(f"Error: Invalid engine(s): {', '.join(invalid)}")
+                print("Use `optics setup --list` to see available engines")
                 return
-            requirements = []
-            for driver in driver_to_install:
-                requirements.extend(ALL_DRIVERS[driver].packages)
-            install_packages(requirements)
+            install_extras(engines)
         else:
-            DriverInstallerApp().run()
+            EngineInstallerApp().run()
 
 
 def main():
@@ -415,7 +411,7 @@ def main():
         ExecuteCommand(),
         LiveCommand(),
         GenerateCommand(),
-        DriverInstaller(),
+        EngineInstaller(),
         ServerCommand(),
         MCPCommand(),
         AutocompletionCommand(),
--- a/optics_framework/helper/setup.py
+++ b/optics_framework/helper/setup.py
@@ -68,6 +68,16 @@ ALL_ENGINES: Dict[str, EngineBackend] = {
     **ACTION_DRIVERS.engines, **TEXT_ENGINES.engines, **LLM_ENGINES.engines
 }
 
+# Convenience bundles — mirror the aggregate extras in pyproject.toml so a token
+# like "all" or "web" resolves to the same set of engines the matching extra
+# installs. Keys are already in normalised (`_norm`) form.
+_BUNDLES: Dict[str, List[EngineBackend]] = {
+    "mobile": [ACTION_DRIVERS.engines["Appium"]],
+    "web": [ACTION_DRIVERS.engines["Selenium"], ACTION_DRIVERS.engines["Playwright"]],
+    "vision": list(TEXT_ENGINES.engines.values()),
+    "all": list(ALL_ENGINES.values()),
+}
+
 # Maps a checkbox-id prefix to its engine category.
 _CATEGORY_BY_PREFIX = {"action": ACTION_DRIVERS, "text": TEXT_ENGINES, "llm": LLM_ENGINES}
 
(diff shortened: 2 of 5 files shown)


optics_framework/helper/setup.py
```diff
@@ -1,55 +1,99 @@
...
 ACTION_DRIVERS = DriverCategory(
     name="Action Driver",
     drivers={
-        "Appium": DriverPackage(name="Appium", packages=["appium-python-client"]),
-        "BLE": DriverPackage(name="BLE", packages=["pyserial"]),
-        "Selenium": DriverPackage(name="Selenium", packages=["selenium", "beautifulsoup4"]),
-        "Playwright": DriverPackage(name="Playwright", packages=["playwright"]),
+        "Appium": Driver(name="Appium", extra="appium", packages=["appium-python-client"]),
+        "BLE": Driver(name="BLE", extra="ble", packages=["pyserial"]),
+        "Selenium": Driver(name="Selenium", extra="selenium", packages=["selenium"]),
```
thakur-patel: confirm if beautifulsoup4 is required or not

optics_framework/helper/cli.py
```diff
@@ -361,28 +361,24 @@ def execute(self, args):
...
-            "setup", help="Install driver for the project")
-        parser.add_argument("--install", nargs="+",
-                            help="Install specified drivers")
+            "setup", help="Install optional engine backends (drivers/OCR/LLM)")
+        parser.add_argument(
+            "--install", nargs="+", metavar="NAME",
+            help="Install the given engines, e.g. --install appium easyocr")
         parser.add_argument("--list", action="store_true",
-                        help="List all available drivers")
+                        help="List all available engines")
```
thakur-patel, line 369: have we renamed drivers to engines?

optics_framework/helper/setup.py
```diff
@@ -1,55 +1,99 @@
...
 from pydantic import BaseModel
 
 
-class DriverPackage(BaseModel):
-    name: str
-    packages: List[str]
+DISTRIBUTION_NAME = "optics-framework"
+
+
+class Driver(BaseModel):
```
thakur-patel: will there be a confusion between this Driver class and the drivers we use in engines/ ?

optics_framework/helper/setup.py
```diff
@@ -1,55 +1,99 @@
...
 
 
 class DriverCategory(BaseModel):
     name: str
-    drivers: Dict[str, DriverPackage]
+    drivers: Dict[str, Driver]
 
 
-# Driver definitions
+# Driver definitions. The `extra` matches the pyproject extra name, which in turn
```
thakur-patel: why do we need to add the use of 'extra' only?

chinmayajha: It is a standard pyproject design

optics_framework/helper/setup.py
```diff
@@ -1,56 +1,110 @@
...
-        "Gemini": DriverPackage(name="Gemini", packages=["google-genai"])
+LLM_ENGINES = EngineCategory(
+    name="LLM Engine",
+    engines={
+        "Gemini": EngineBackend(name="Gemini", extra="llm", packages=["google-genai"], aliases=["llm"]),
     }
 )
 
-ALL_DRIVERS = {**ACTION_DRIVERS.drivers, **TEXT_DRIVERS.drivers, **LLM_DRIVERS.drivers}
+ALL_ENGINES: Dict[str, EngineBackend] = {
```
kun-codes, line 67: `ALL_ENGINES` is made from the 3 dictionaries. Neither of them entries for mobile/web/vision/all. So despite these bundles existing in `pyproject.toml` they would be rejected by `setup.py`

optics_framework/helper/setup.py
```diff
@@ -91,79 +145,76 @@ def compose(self) -> ComposeResult:
...
-            print("Installing Playwright Chromium driver along with system dependencies...")
-            # have to run this command because as per docs
-            # https://playwright.dev/python/docs/browsers#install-system-dependencies
-            # this separate command has to be run after installing playwright via pip
-            #
-            # used chromium because it is the most popular browser
-            # not specifying "--with-deps chromium" would install all browsers without system
-            # dependencies
+            [sys.executable, "-m", "pip", "install", spec],
+            check=True, shell=False)
```
kun-codes: You should add `capture_output=True, text=True` here so that https://github.com/chinmayajha/optics-framework/blob/788c75ee81a6b9466e2fd2cd9d13087b58778597/optics_framework/helper/setup.py#L211 can print out the error in more detail in case the command fails

optics_framework/helper/setup.py
```diff
@@ -1,56 +1,110 @@
...
+    index: Dict[str, EngineBackend] = {}
+    for engine in ALL_ENGINES.values():
+        for token in [engine.name, engine.extra, *engine.aliases]:
+            index[_norm(token)] = engine
+    return index
 
-# Maps a checkbox-id prefix to its driver category.
-_CATEGORY_BY_PREFIX = {"action": ACTION_DRIVERS, "text": TEXT_DRIVERS, "llm": LLM_DRIVERS}
 
+def resolve_engines(tokens: List[str]) -> tuple[List[EngineBackend], List[str]]:
```
kun-codes, line 100: we should add unit tests for `resolve_engines`, `_alias_index`, `_norm`, `install_extras`

kun-codes (changes requested):

kun-codes:
I feel using the terms engines and drivers is better since drivers is used for the important things like playwright, selenium, appium etc while engines are for accessories like llm, easyocr etc. 

Although maybe we can rename engines to helpers

kun-codes:
I tested out this pr locally and it works fine. I didn't do a thorough testing though

optics_framework/helper/setup.py
chinmayajha: Fixed in de6269b — resolve_engines now expands the mobile/web/vision/all bundles to their engines, and list_engines prints them too.

optics_framework/helper/setup.py
chinmayajha: Done in de6269b — both the pip and playwright runs now use capture_output=True, text=True, and the except prints e.stderr (falling back to stdout).

optics_framework/helper/setup.py
chinmayajha: Added in 639c4fa — tests/units/helpers/test_setup.py covers _norm, _alias_index, resolve_engines and install_extras, plus a pyproject parity guard.

pyproject.toml
```diff
@@ -45,14 +45,43 @@ scikit-image = "^0.25.2"
 uvicorn = ">=0.35,<0.43"
 jsonpath-ng = "^1.7.0"
 beautifulsoup4 = "^4.12"
+# Optional engine backends. These are imported by the drivers / element sources /
+# vision models under optics_framework/engines/, but the core framework runs
+# without them — install the matching extra (or run `optics setup`) to enable an
+# engine. Version bounds are the minimum-tested floors with a next-major ceiling.
```
malto101: remove unnecessary explanation

malto101 (approved):
Looks good, tested it out, works fine

pyproject.toml
chinmayajha: fixed

kun-codes (approved):

Issue #392, opened 2026-07-24
Reconcile global vs project config (optics config vs optics execute)


chinmayajha:
Deferred from the onboarding overhaul (PR to follow).

## Problem
`optics config` edits the **global** file `~/.optics/global_config.yaml`, but `optics execute <folder>`/`optics dry_run <folder>` read **only** that folder's own `config.yaml` and never merge the global file (the runner assigns `self.config = config_obj` directly and does not call `ConfigHandler.load()`). So editing config via the TUI has no effect on execution — a confusing, disjoint model.

The onboarding docs now document this accurately and the config-TUI shows which file it edits, but the underlying model is unchanged.

## Options
1. Make `optics config` operate on a project's `config.yaml` (project-first model), or
2. Have `optics execute`/`dry_run` merge `~/.optics/global_config.yaml` under the project config so global edits act as defaults.

## Refs
- `optics_framework/helper/config_manager.py` (global-only save)
- `optics_framework/helper/execute.py` (reads project config directly)
- `optics_framework/common/config_handler.py` (`load()`/`deep_merge` exist but the runner doesn't use them)

chinmayajha:
Outdated — resolved by PR #467 (project-scoped configure). `optics config` is deprecated and forwards to project-scoped `configure` (`cli.py`), the TUI edits the project's own `config.yaml`, and the global `~/.optics/global_config.yaml` has been removed (`config_handler.py`) — the disjoint global-vs-project model no longer exists. Closing.

PR #403, opened 2026-07-24
fix(init,config,completion): valid scaffold, stop dep-list corruption, refresh completions

chinmayajha:
Part **3 of 4** of the onboarding overhaul split out of #395. This chunk fixes first-run UX: `optics init` scaffolding, the `optics config` TUI dependency-list corruption, and stale shell completions.

## `fix(init)`: produce a valid, runnable project scaffold
`optics init` previously built `config.yaml` by reading `~/.optics/global_config.yaml`, which does not ship and does not exist on a fresh machine — so it wrote a config containing literally `''` while printing "Created config.yaml with default values." It also always wrote flat root CSVs and *then* copied any template on top, yielding a confusing hybrid layout, and copied `.DS_Store` into the project.
- **No template:** scaffold the sample subdir layout (`test_cases/`, `modules/`, `test_data/`) plus a commented starter `config.yaml` mirroring framework defaults (every source present but disabled, documented how to enable one). The result is recognised as a valid project config.
- **With a template:** copy the sample verbatim only (no more hybrid), skipping hidden files and `__pycache__`.
- Unknown template now lists available templates; fix the `--template` help text (was the nonexistent `sample1`); print next-step guidance.

## `fix(config)`: stop corrupting dependency lists
The config TUI keyed every dependency-list entry (`driver_sources`, `elements_sources`, …) by the literal string `"name"`, collapsing e.g. `['appium']` to `[{'name': ...}]` — an unresolvable config. Key each entry by its actual source name. Also show which file the TUI edits (the GLOBAL `~/.optics` config) and note that `optics execute <folder>` reads the folder's own `config.yaml` (see #392 for reconciling the two surfaces).

## `fix(completion)`: refresh shell completions
The embedded scripts advertised a nonexistent `version` subcommand and omitted `live`/`mcp`; the template list missed `playwright` (and stale top-level copies had a `calender` typo and missed `clock`). Fix the subcommand/template lists, add `live`/`mcp`, adjust the driver-list filter to the new `setup --list` output, regenerate the top-level copies, and add default cases to every shell `case` statement (SonarQube).

## Verification
Full suite green on top of current `main`: **565 passed, 1 skipped, 2 xfailed**.

Relates to #392 (global vs project config reconciliation).

---
### Completion ↔ setup coupling (added commit)
`optics setup --list` (#402) renamed its category headers to "OCR Engines:" / "LLM Engines:". The completion scripts scraped keys by blacklisting header words — fragile, and it leaked any un-blacklisted header. Now they parse **only the two-space-indented key lines**, which is header-agnostic (also fixes a pre-existing leak), and the remaining `driver`→`engine` labels/vars are aligned with setup's vocabulary. Top-level `autocompletion/` copies regenerated; both pass `bash -n`/`zsh -n`.

### DIFF
--- a/autocompletion/optics_completion.sh
+++ b/autocompletion/optics_completion.sh
@@ -6,7 +6,7 @@ _optics_completions() {
 
   local subcommands="list config dry_run init execute live generate setup serve mcp completion"
 
-  local template_options="contact clock calendar youtube gmail_web playwright"
+  local template_options="calendar clock contact gmail_web playwright youtube"
   local runner_options="test_runner pytest"
   # Only the two-space-indented lines are engine keys; category headers are not
   # indented, so this needs no per-header filter.
--- a/autocompletion/optics_completion.sh
+++ b/autocompletion/optics_completion.sh
@@ -1,27 +1,37 @@
 #!/bin/bash
 
 _optics_completions() {
-  local cur prev
+  local cur prev words cword
   _init_completion || return
 
-  local subcommands="list config dry_run init execute version generate setup serve completion"
+  local subcommands="list config dry_run init execute live generate setup serve mcp completion"
 
-  local template_options="calender contact gmail_web youtube"
+  local template_options="contact clock calendar youtube gmail_web playwright"
   local runner_options="test_runner pytest"
-  local driver_options=$(optics setup --list 2>/dev/null | awk '{print $1}' | grep -vE '^(Action|Available|Drivers:|Text)$')
+  local driver_options=$(optics setup --list 2>/dev/null | awk '{print $1}' | grep -vE '^(Action|Available|Text|LLM)$')
 
   case ${COMP_CWORD} in
     1)
       COMPREPLY=( $(compgen -W "$subcommands" -- "$cur") )
       return 0
       ;;
-    *)
-      ;;
   esac
 
   case ${COMP_WORDS[1]} in
-    list|config|version|completion)
+    list|config|completion)
       COMPREPLY=( $(compgen -W "--help -h" -- "$cur") )
+      ;;
+
+    live)
+      COMPREPLY=( $(compgen -d -- "$cur") )
+      ;;
+
+    mcp)
+      if [[ $prev == "--transport" ]]; then
+        COMPREPLY=( $(compgen -W "stdio http" -- "$cur") )
+      else
+        COMPREPLY=( $(compgen -W "--transport --host --port -h --help" -- "$cur") )
+      fi
       ;;
 
     dry_run|execute)
@@ -45,13 +55,13 @@ _optics_completions() {
 
     generate)
       COMPREPLY=( $(compgen -W "-h --help" -- "$cur") )
-      if [[ $prev == "--framework" ]]; then
-        COMPREPLY=( $(compgen -W "pytest robot" -- "$cur") )
-      elif [[ $prev == "--output" ]]; then
-        COMPREPLY=( $(compgen -f -- "$cur") )
-      elif [[ $prev == "--project_path" ]]; then
-        COMPREPLY=( $(compgen -d -- "$cur") )
-      fi
+        if [[ $prev == "--framework" ]]; then
+            COMPREPLY=( $(compgen -W "pytest robot" -- "$cur") )
+        elif [[ $prev == "--output" ]]; then
+            COMPREPLY=( $(compgen -f -- "$cur") )
+        elif [[ $prev == "--project_path" ]]; then
+            COMPREPLY=( $(compgen -d -- "$cur") )
+        fi
       ;;
 
     setup)
@@ -67,10 +77,6 @@ _optics_completions() {
 
     serve)
       COMPREPLY=( $(compgen -W "--host --port -h --help" -- "$cur") )
-      ;;
-
-    *)
-      COMPREPLY=()
       ;;
   esac
 }
--- a/autocompletion/optics_completion.sh
+++ b/autocompletion/optics_completion.sh
@@ -15,6 +15,9 @@ _optics_completions() {
       COMPREPLY=( $(compgen -W "$subcommands" -- "$cur") )
       return 0
       ;;
+    *)
+      # Beyond the subcommand: handled per-subcommand below.
+      ;;
   esac
 
   case ${COMP_WORDS[1]} in
@@ -61,6 +64,8 @@ _optics_completions() {
             COMPREPLY=( $(compgen -f -- "$cur") )
         elif [[ $prev == "--project_path" ]]; then
             COMPREPLY=( $(compgen -d -- "$cur") )
+        else
+            : # keep the default --help completion set above
         fi
       ;;
 
@@ -77,6 +82,11 @@ _optics_completions() {
 
     serve)
       COMPREPLY=( $(compgen -W "--host --port -h --help" -- "$cur") )
+      ;;
+
+    *)
+      # Unknown subcommand: offer only --help.
+      COMPREPLY=( $(compgen -W "-h --help" -- "$cur") )
       ;;
   esac
 }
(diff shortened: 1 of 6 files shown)


optics_framework/helper/initialize.py
```diff
@@ -1,9 +1,84 @@
...
+  - appium_page_source:
+      enabled: false
+  - appium_screenshot:
+      enabled: false
+  # Playwright locators / page source / screenshots:
+  - playwright_find_element:
+      enabled: false
+  - playwright_page_source:
+      enabled: false
+  - playwright_screenshot:
```
kun-codes, line 91: why don't we have something like this for selenium?

`selenium_find_element`, `selenium_page_source`, `selenium_screenshot`

optics_framework/helper/autocompletion.py
```diff
@@ -9,18 +9,23 @@
...
-local -a templates=("calendar" "clock" "contact" "gmail_web" "youtube")
+local -a templates=("contact" "clock" "calendar" "youtube" "gmail_web" "playwright")
 local -a runners=("test_runner" "pytest")
 local -a frameworks=("pytest" "robot")
-local -a drivers=("${(f)$(optics setup --list 2>/dev/null | awk '{print $1}' | grep -vE '^(Action|Available|Drivers:|Text)$')}")
+local -a transports=("stdio" "http")
+# `optics setup --list` indents each installable engine key by two spaces under a
+# category header; take only those indented lines so this stays correct no matter
+# what the category headers are named.
+local -a engines=("${(f)$(optics setup --list 2>/dev/null | awk '/^  / {print $1}')}")
```
kun-codes, line 33: `awk '/^  / {print $1}')` will split multi word drivers by spaces and only return the first word of the driver's name. So autocomplete suggests invalid name for "Google Vision"

optics_framework/helper/autocompletion.py
kun-codes: have you tested this on a zsh shell?

optics_framework/helper/cli.py
```diff
@@ -211,7 +211,9 @@ def register(self, subparsers: argparse._SubParsersAction):
             "--force", action="store_true", help="Override if the project exists"
         )
         parser.add_argument(
-            "--template", help="Select a project template (e.g., 'sample1')"
+            "--template",
+            help="Start from a sample template: contact, clock, calendar, "
```
kun-codes: this, https://github.com/chinmayajha/optics-framework/blob/1a8536dc34c384c0933cd8d301f1e3003d7a7194/optics_framework/helper/autocompletion.py#L21  and https://github.com/chinmayajha/optics-framework/blob/1a8536dc34c384c0933cd8d301f1e3003d7a7194/optics_framework/helper/autocompletion.py#L125

hardcode the names of the samples but https://github.com/chinmayajha/optics-framework/blob/1a8536dc34c384c0933cd8d301f1e3003d7a7194/optics_framework/helper/initialize.py#L124 computes it at run time

kun-codes (changes requested):

optics_framework/helper/initialize.py
chinmayajha: Good catch — fixed in 6f76490. Added the selenium_find_element / selenium_page_source / selenium_screenshot block to the starter config (all three exist under engines/elementsources/), mirroring appium and playwright.

optics_framework/helper/autocompletion.py
chinmayajha: This parsing targets the #402 `setup --list` output, where each installable token is a single-word extra key (appium, google-vision, easyocr, plus the bundles mobile/web/vision/all) — no spaces, so $1 is exactly the token. What you're seeing is the pre-402 `list_drivers` output still on this branch, which prints display names ("Google Vision"). This branch's completion depends on #402; it should land/rebase after it (noted in the PR body's coupling section). Once #402 is in, $1 is correct.

optics_framework/helper/autocompletion.py
chinmayajha: Yes. zsh -n passes, and loading it under compinit registers _optics_completions and the arrays populate correctly. Both the embedded script and the top-level autocompletion/ copy pass zsh -n.

optics_framework/helper/cli.py
chinmayajha: Fixed in bb0d3f3. Added initialize.available_templates() as the single source of truth; the --template help now builds its list from it, and the completion scripts carry placeholders that write_completion_scripts() fills from the same function. No more hardcoded lists to drift.

autocompletion/optics_completion.sh
```diff
@@ -1,29 +1,44 @@
 #!/bin/bash
 
 _optics_completions() {
-  local cur prev
+  local cur prev words cword
   _init_completion || return
 
-  local subcommands="list config dry_run init execute version generate setup serve completion"
+  local subcommands="list config dry_run init execute live generate setup serve mcp completion"
```
malto101, line 7: any reason for removing version?

autocompletion/optics_completion.zsh
```diff
@@ -1,82 +1,82 @@
...
+    'dry_run: Run tests in dry-run mode'
+    'init: Initialize a new project'
+    'execute: Execute tests'
+    'live: Interactive session against a live target'
+    'generate: Generate framework code'
+    'setup: Install optional engine backends'
+    'serve: Start the optics REST server'
+    'mcp: Start the MCP server'
+    'completion: Enable shell completion'
 )
```
malto101: I think I missed it earlier, could you add helper cmd in bash if possible

autocompletion/optics_completion.zsh
```diff
@@ -1,82 +1,82 @@
...
-            '--help[-h]'
-          ;;
+                live)
+                    _arguments                         '*:project_path:_files -/'                         '--help[-h]'
+                    ;;
 
-        init)
-          _arguments \
-            '--name=[Project name]' \
-            '--path=[Project path]' \
```
malto101, line 49: feels harder to read, if its not reline
and unnecessary diff is being generated

malto101 (changes requested):
small nitpick

autocompletion/optics_completion.sh
chinmayajha: There's no `version` subcommand, version is the top-level `--version` flag

autocompletion/optics_completion.zsh
chinmayajha: The descriptions in the zsh script come from `_describe`. In bash, `compgen -W` completes the literal word, so there's no clean way to attach `cmd: description` labels, they'd get inserted as the actual completion text. 

so the bash script stays names-only by design, as it already lists every subcommand and offers `-h/--help` per command. If you had a specific helper in mind, pls let me know.

autocompletion/optics_completion.zsh
chinmayajha: Fixed in 7739b0b. Root cause: the zsh template was a plain Python string, so every `\`+newline line-continuation got collapsed at import time — that's why the `_arguments` blocks came out as one long space-padded line. Made the template a raw string and kept the original 2-space indentation, so it's back to readable multi-line and the diff is down to just the real changes (version removed, live/mcp added, engine parsing, default cases). Regenerated the top-level copies; both pass `zsh -n`/`bash -n`.

autocompletion/optics_completion.sh
malto101: Then I would suggest to keep --version

chinmayajha: Added

chinmayajha: Fair point — fixed in e18aad0. `--version` (and `--help`) were never offered at position 1 even before this PR, since the old `version` entry was standing in for the flag but wasn't wired as a real completion option either. Added both as proper top-level completions:

- bash: `COMP_CWORD==1` now offers `$subcommands --version --help -h`
- zsh: declared via `(-)--version[...]` / `(-)--help[-h]` in the top-level `_arguments -C` so they're recognised as flags, not command names

Regenerated the top-level `autocompletion/` copies; both pass `bash -n`/`zsh -n`.

kun-codes (approved):

chinmayajha:
@malto101 Is this alright now?

malto101 (approved):
👌

Issue #394, opened 2026-07-24
Clean up sample/doc test-data inconsistencies (gmail_web elements, elements.csv columns)


chinmayajha:
Deferred from the onboarding overhaul (PR to follow).

## Problems
1. **gmail_web sample** `test_data/elements.csv` contains YouTube element names (`Shorts`, `Subscriptions`, `Today button`, etc.) that don't match the Gmail flow in its `modules.csv` — stale/copied data.
2. **elements.csv column format mismatch:** `docs/quickstart.md` documents a multi-column format (`Element_Name,Element_ID_xpath,Element_ID,Element_Text`), but every sample and `optics init` use `Element_Name,Element_ID`. Need to confirm which columns the CSV reader (`common/runner/data_reader.py`) actually supports and reconcile the docs to it.

## Refs
- `optics_framework/samples/gmail_web/test_data/elements.csv`
- `docs/quickstart.md` (Step: Define Your Elements)
- `optics_framework/common/runner/data_reader.py` (element CSV parsing)

malto101:
@chinmayajha can we close this now?

chinmayajha:
> [@chinmayajha](https://github.com/chinmayajha) can we close this now?

yess

Issue #393, opened 2026-07-24
Surface Docker & devcontainer in the docs site; fix broken Docker link


chinmayajha:
Deferred from the onboarding overhaul (PR to follow).

## Problem
- The `Docker/` deployment guide (REST + MCP images) and `.devcontainer/` (a working dev environment) are **not referenced** anywhere in the mkdocs site or README, so newcomers never discover them.
- `docs/usage/mcp_usage.md` links to `../../Docker/deployment.md`, which mkdocs cannot resolve because `Docker/` lives outside the `docs/` tree (build warning: target `../Docker/deployment.md` not found).

## Suggested fix
- Add a 'Deployment (Docker)' page and a short 'Dev environment / Codespaces' note to the mkdocs nav (either move the content under `docs/` or include it), and fix the mcp_usage link.

## Refs
- `Docker/deployment.md`, `docker-compose.yml`, `.devcontainer/devcontainer.json`
- `docs/usage/mcp_usage.md` (broken link)
- `mkdocs.yml` (nav)

malto101:
what about status on this?

chinmayajha:
> what about status on this?

will work on this (and other issues) from next week

PR #404, opened 2026-07-24
docs: overhaul onboarding to match actual CLI behaviour

chinmayajha:
Part **4 of 4** of the onboarding overhaul split out of #395. Docs only — rewrites onboarding to match actual CLI behaviour (best reviewed/merged after the setup-extras and init/config chunks, since it documents that behaviour).

## What changed
- **New Installation & Prerequisites page** (`docs/prerequisites.md`): Python 3.12+, the extras table, `optics setup`, and the external tooling each platform needs (Appium server, Node, Android SDK/adb, JDK, browsers, OCR binaries). Added to the mkdocs nav between Introduction and Quick Start.
- **README:** describe the extras-based install; fix the Quick Start flow (install engines → `init --template` → `dry_run` → `execute`); correct `optics version` → `optics --version`; fix the stale developer section (MkDocs not Sphinx, ruff/pre-commit not Black, `poetry install --with dev,test`, correct directory/package tree; drop the nonexistent `dev_requirements.txt`).
- **Quick Start / index:** use config-key setup names, add the `dry_run` step, fix the project-location diagrams (created in the CWD, not under `optics_framework/samples/`).
- **CLI Usage:** rewrite the setup section for extras; correct `optics config` (interactive TUI, no `--set/--reset/--list` flags); add `playwright` to the template list.
- **Configuration:** correct the load model — `optics execute`/`dry_run` read the project `config.yaml` and do not merge the global `~/.optics` config; document which surface each command touches.
- **User Workflow:** reconcile the project layout with the subdir scaffold, rename `test_modules.csv` → `modules.csv`, de-duplicate the `dry_run` example.

## Verification
`mkdocs build` succeeds; full suite green on top of current `main`: **565 passed, 1 skipped, 2 xfailed**.

Relates to #392 (config model), #393 (Docker/devcontainer docs + broken mcp link), #394 (sample/doc test-data inconsistencies).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,47 +31,45 @@ This framework was designed primarily for the following use cases:
 
 ## 📦 Installation
 
-### Install via `pip`
+Optics needs **Python 3.12+**. Install the core CLI with `pip`:
 
 ```bash
 pip install optics-framework
 ```
+
+The core install is deliberately light — the drivers, OCR engines and LLM backends are **optional extras** you add per project. Install only what you need:
+
+```bash
+pip install "optics-framework[appium]"       # native Android/iOS via Appium
+pip install "optics-framework[playwright]"   # browser automation via Playwright
+pip install "optics-framework[easyocr]"      # on-screen text detection (OCR)
+```
+
+Available extras: `appium`, `selenium`, `playwright`, `ble`, `easyocr`, `pytesseract`, `google-vision`, `llm`, `mcp`, plus bundles `mobile`, `web`, `vision`, and `all`. You can also install them interactively with `optics setup` (see below).
+
+> A driver extra installs only the **Python client**. For mobile testing you also need the **Appium server**, a device/emulator, and platform tooling (Node.js, Android SDK/adb, JDK). See the [Installation & Prerequisites guide](https://mozarkai.github.io/optics-framework/prerequisites/).
+
+> **⚠️ Conda note:** `easyocr` and `optics-framework` conflict under Conda (numpy 1.x vs 2.x). Use a standard `venv` instead.
 
 ---
 
 ## 🚀 Quick Start
 
-### 1 Install Optics Framework
-
-**Note**: Ensure Appium server is running and a virtual Android device is enabled before proceeding.
-
 ```bash
-mkdir ~/test-code
-cd ~/test-code
-python3 -m venv venv
-source venv/bin/activate
+# 1. Create an isolated environment and install Optics
+python3 -m venv venv && source venv/bin/activate
 pip install optics-framework
-```
 
-> **⚠️ Important:** Conda environments are not supported for `easyocr` and `optics-framework` together, due to conflicting requirements for `numpy` (version 1.x vs 2.x). Please use a standard Python virtual environment instead.
+# 2. Install the engines you need (equivalent to the [appium]/[easyocr] extras)
+optics setup --install appium easyocr
 
-### 2 Create a New Test Project
+# 3. Scaffold a project from a ready-made sample
+optics init --name my_test_project --template contact
 
-```bash
-optics setup --install Appium EasyOCR
-optics init --name my_test_project --path . --template contact
-```
-
-### 📌 Dry Run Test Cases
-
-```bash
-optics dry_run my_test_project
-```
-
-### 📌 Execute Test Cases
-
-```bash
-optics execute my_test_project
+# 4. Point config.yaml at your device/app (see the sample's config.yaml),
+#    make sure the Appium server + emulator are running, then:
+optics dry_run my_test_project      # validate the suite without a device
+optics execute my_test_project      # run it for real
 ```
 
 ---
@@ -87,7 +85,7 @@ optics execute <project_name>
 ### Initialize a New Project
 
 ```bash
-optics init --name <project_name> --path <directory> --template <contact/youtube> --force
+optics init --name <project_name> --path <directory> --template <contact|youtube|...> --force
 ```
 
 ### List Available Keywords
@@ -105,37 +103,12 @@ optics --help
 ### Check Version
 
 ```bash
-optics version
+optics --version
 ```
 
 ---
 
 ## 🏗️ Developer Guide
-
-### Project Structure
-
-```bash
-Optics_Framework/
-├── LICENSE
-├── README.md
-├── dev_requirements.txt
-├── samples/            # Sample test cases and configurations
-|   ├── contact/
-|   ├── youtube/
-├── pyproject.toml
-├── tox.ini
-├── docs/               # Documentation using Sphinx
-├── optics_framework/   # Main package
-│   ├── api/            # Core API modules
-│   ├── common/         # Factories, interfaces, and utilities
-│   ├── engines/        # Engine implementations (drivers, vision models, screenshot tools)
-│   ├── helper/         # Configuration management
-├── tests/              # Unit tests and test assets
-│   ├── assets/         # Sample images for testing
-│   ├── units/          # Unit tests organized by module
-│   ├── functional/     # Functional tests organized by module
-
-```
 
 ### Available Keywords
 
@@ -326,15 +299,18 @@ The following keywords are available and organized by category. These keywords c
 
 ```bash
 git clone git@github.com:mozarkai/optics-framework.git
-cd Optics_Framework
+cd optics-framework
 pipx install poetry
-poetry install --with dev
+poetry install --with dev,test
+poetry run pre-commit install
 ```
+
+To work on an engine backend, add its extra, e.g. `poetry install -E appium -E easyocr`.
 
 ### Running Tests
 
 ```bash
-poetry install --with tests
+poetry install --with test
 poetry run pytest
 ```
 
@@ -362,7 +338,7 @@ We welcome contributions! Please follow these steps:
 3. Commit your changes.
 4. Open a pull request.
 
-Ensure your code follows **PEP8** standards and is formatted with **Black**.
+Ensure your code follows **PEP8** and passes the pre-commit hooks (ruff lint + format, bandit, commitizen). Run `poetry run pre-commit run --all-files` before pushing.
 
 ---
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,14 +4,13 @@ This document provides comprehensive documentation for all configuration options
 
 ## Overview
 
-The Optics Framework uses a hierarchical configuration system that merges configurations in the following order (with later configurations taking priority):
+There are two configuration files:
 
-1. **Default Configuration** - Built-in defaults
-2. **Global Configuration** - `~/.optics/global_config.yaml` (user-wide settings)
-3. **Project Configuration** - `config.yaml` in your project directory (project-specific settings)
+1. **Project Configuration** — `config.yaml` in your project directory. This is the config `optics execute <folder>` / `optics dry_run <folder>` actually read. Built-in defaults fill in any field you omit.
+2. **Global Configuration** — `~/.optics/global_config.yaml`, edited by the interactive `optics config` command.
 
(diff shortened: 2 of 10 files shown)


docs/prerequisites.md
```diff
@@ -0,0 +1,135 @@
+# Installation & Prerequisites
```
malto101, line 1: I dont understand the meaning behind this page?
isnt this a amalgamation of QuickStart and configuration files

docs/usage/CLI_usage.md
```diff
@@ -140,17 +148,19 @@ optics --help
...
+`q` to quit). It takes no flags.
 
-- `--set <key> <value>`: Set a configuration key-value pair.
-- `--reset`: Reset all configurations to default.
-- `--list`: Display current configuration values.
+!!! warning "Global vs per-project config"
+    `optics config` edits the **global** file. `optics execute <folder>` reads
+    that folder's **own** `config.yaml`, which does **not** merge the global
+    file. For a specific project, edit its `config.yaml` directly — that is the
+    config the runner uses.
```
malto101: lets wait for the optics config fix to merge and then remove this warning

docs/configuration.md
```diff
@@ -4,14 +4,20 @@ This document provides comprehensive documentation for all configuration options
 
 ## Overview
 
-The Optics Framework uses a hierarchical configuration system that merges configurations in the following order (with later configurations taking priority):
+There are two configuration files:
```
malto101, line 7: ideally what was originally written should be maintained 
if it isnt right, that's mean it should be fixed

docs/configuration.md
```diff
@@ -117,7 +123,10 @@ All configurations are defined in YAML format. The main configuration file (`con
 
     **Type:** `Optional[str]` | **Default:** `null`
 
-    Root directory for test project files. This path should contain your CSV files (`test_cases.csv`, `test_modules.csv`, `elements.csv`) and `input_templates/` directory.
+    Root directory for test project files. This folder holds your test cases,
+    modules, and elements (either the sample subdir layout — `test_cases/`,
+    `modules/`, `test_data/` — or flat CSVs; the runner discovers them by
+    content) plus any `input_templates/` images.
```
malto101: why 4 new line?

docs/configuration.md
```diff
@@ -823,46 +832,20 @@ driver_sources:
...
-        deviceName: "emulator-5554"
-```
-
-**Result:** The merged configuration will have `log_level: DEBUG` and the Appium capabilities will include both `platformName: "Android"` and `deviceName: "emulator-5554"`.
+!!! note "Where the global config is used"
+    The default → global → project deep-merge is performed by `ConfigHandler.load()`,
+    which the interactive `optics config` command uses to edit the **global**
+    file. The batch runner does not call it, so editing the global config has no
+    effect on `optics execute`. To change a specific run, edit that project's
+    `config.yaml`.
```
malto101: same here, if this is the case, it means its broken, also usage of technical apis like this is not forbidden per say but highly discouraged as our audience is for people who haven't touch code

docs/quickstart.md
```diff
@@ -2,21 +2,15 @@
...
-
-### Install Optics Framework
-
-```bash
-pip install optics-framework
-```
+!!! warning "Prerequisites"
+    This guide assumes you have already installed Optics and the platform tooling
+    for your target (for Android: the Appium server, an emulator/device, adb).
+    If not, do the [Installation & Prerequisites](prerequisites.md) guide first.
```
malto101: a QuickStart guide is always a entry, people first enter through QuickStart, the idea is for the user to test out the product with bare min understanding and involvement

docs/quickstart.md
```diff
@@ -244,6 +248,9 @@ Test cases define **what** to test, modules define **how** to do it. This separa
 - Appium server running: `appium`
 - Android virtual device or physical device connected and verified: `adb devices`
 
+See [Installation & Prerequisites](prerequisites.md) for how to install the
+Appium server, Android SDK/adb, and a JDK.
```
malto101: I would suggest instead of having us maintain the guide for installing 3rd party SBOM, we just share the link of installation of their docs, because everyone has their way of installation

docs/user_workflow.md
```diff
@@ -4,21 +4,23 @@ This guide walks you through the workflow for using the Optics Framework to crea
...
-- **--path**: Defines where the project folder will be created.
-- **--template**: (Optional) Uses a predefined sample template to populate initial files.
+- **--name**: Specifies the project name (required).
+- **--path**: Directory where the project folder is created (default: current directory).
+- **--template**: (Optional) Copy a complete sample project (`contact`, `clock`, `calendar`, `youtube`, `gmail_web`, `playwright`).
 
-This command generates a project directory with a default structure, including a `config.yaml` and placeholder CSV files.
+With `--template`, the project is a ready-to-run copy of that sample. Without it,
+`optics init` scaffolds the subdirectory layout (`test_cases/`, `modules/`,
+`test_data/`) plus a commented starter `config.yaml` for you to fill in.
```
malto101: current templates only use csv, we don't have yaml style templates, would be a nice addition 

plus I remember the parser for yaml in data_reader was bad, need to revisit that

docs/user_workflow.md
```diff
@@ -160,25 +162,33 @@ exclude:
...
 ```
 
-- **input_templates/**: Store any input images (e.g., `home.png`) referenced in `elements.csv`.
+- **test_data/input_templates/**: Store any input images (e.g., `home.png`) referenced in `elements.csv`.
 - **execution_output/**: Contains logs and other output generated during test runs.
 
+!!! note "File discovery is content-based"
+    The runner identifies each file by its contents, not its name or location, so
+    a flat layout (all CSVs in the project root) also works. The subdirectory
+    layout above is what the samples and `optics init` use.
```
malto101: content is good, but unnecessary new line here

docs/user_workflow.md
```diff
@@ -207,13 +217,9 @@ To check for syntactical errors in your CSV files and configuration, run a dry r
...
-For a specific test case:
-
-```bash
-optics dry_run ./contact
-```
-
-This command simulates the test execution without interacting with the device, helping you catch issues early.
+`dry_run` takes a project folder (not a single test case). It simulates
+execution without interacting with the device — validating that every keyword
+exists and every `${element}` resolves — helping you catch issues early.
```
malto101: once again with the new line, also would suggest to keep the cmd back, content is good

README.md
```diff
@@ -31,49 +31,60 @@ This framework was designed primarily for the following use cases:
...
+#    make sure the Appium server + emulator are running, then:
+optics dry_run my_test_project      # validate the suite without a device
+optics execute my_test_project      # run it for real
 ```
 
+`optics init` without `--template` scaffolds an empty project
+(`test_cases/`, `modules/`, `test_data/`, and a commented `config.yaml`) for you
+to fill in. Templates: `contact`, `clock`, `calendar`, `youtube`, `gmail_web`,
+`playwright`.
+
```
malto101: I dont think this is needed information for people just trying it out for the first time

README.md
```diff
@@ -115,26 +126,22 @@ optics version
...
+├── docs/                   # Documentation (MkDocs Material)
+├── optics_framework/       # Main package
+│   ├── api/                # Keyword classes (ActionKeyword, Verifier, ...)
+│   ├── common/             # Factories, interfaces, execution engine, utilities
+│   ├── engines/            # Drivers, element sources, vision models, LLMs
+│   ├── helper/             # CLI, project/init/setup/config helpers
+│   └── samples/            # Sample projects used by `optics init --template`
+└── tests/                  # Unit + feature tests
+    ├── units/
+    └── feature/
```
malto101: I so think we should just remove from README,
I mean we literally see the dir in GitHub when first open the repo

README.md
```diff
@@ -326,15 +333,17 @@ The following keywords are available and organized by category. These keywords c
...
 ```
 
+To work on an engine backend, add its extra, e.g. `poetry install -E appium -E easyocr`.
+
 ### Running Tests
 
 ```bash
-poetry install --with tests
 poetry run pytest
 ```
```
malto101, line 315: as its a part of separate heading from above,
I would suggest to treat it separate

that's why could you bring back the install dep for tests

README.md
```diff
@@ -362,7 +371,8 @@ We welcome contributions! Please follow these steps:
 3. Commit your changes.
 4. Open a pull request.
 
-Ensure your code follows **PEP8** standards and is formatted with **Black**.
+Ensure your code passes the pre-commit hooks (ruff lint + format, bandit,
+commitizen). Run `poetry run pre-commit run --all-files` before pushing.
```
malto101: this is good, but ideally would like everything under PEP8 standard,
neat and easy to look at.
plus everyone can set it on style

malto101 (changes requested):
the one line were weird

docs/prerequisites.md
chinmayajha: Fair point. I trimmed this so it no longer repeats QuickStart or the config page. What is left is the extras table, which does not live anywhere else, plus links to each tool's own install docs. QuickStart no longer depends on this page.

docs/usage/CLI_usage.md
chinmayajha: Agreed. #403 adds this same note inside the optics config TUI, so the warning here is a duplicate. I removed it since #403 is landing soon.

docs/configuration.md
chinmayajha: The old text described a merge order that only the optics config command actually does. The runner reads the project config directly, so that wording was wrong for execute and dry_run. I kept the fix but wrote it in plainer language. The behaviour itself is worth a code fix, so I opened #409 to track it.

docs/configuration.md
chinmayajha: That was leftover line wrapping. I put it back on one line to match the rest of the docs.

docs/configuration.md
chinmayajha: I removed the ConfigHandler.load() and Config references and wrote it in plain terms instead. I agree the behaviour looks broken, so I opened #409 to track the code fix separately from the docs.

docs/quickstart.md
chinmayajha: Agreed. I removed the prerequisites gate, so QuickStart and the landing page are the entry point again with install shown inline.

docs/quickstart.md
chinmayajha: Done. I replaced our own steps with links to the Appium and Android platform tools docs. The prerequisites page also links out for every external tool now.

docs/user_workflow.md
chinmayajha: Good point. The docs do not claim YAML templates, so there is nothing to change here. I will treat YAML templates and the data_reader YAML parser as a separate follow up rather than mixing them into a docs PR.

docs/user_workflow.md
chinmayajha: Fixed the line wrapping here.

docs/user_workflow.md
chinmayajha: Fixed the wrapping, and I kept the dry_run command block above the text.

README.md
chinmayajha: Agreed. I removed that paragraph. The template list is still in the Usage and CLI sections where people actually need it.

README.md
chinmayajha: Done. I removed the project structure tree.

README.md
chinmayajha: I put poetry install --with test back under Running Tests so the section works on its own.

README.md
chinmayajha: I kept the PEP8 line and added the pre-commit note next to it.

malto101 (approved):
LGTM

PR #408, opened 2026-07-24
fix: pin and wheel-only-restrict pip/poetry installs across CI and Docker

chinmayajha:
## Summary

Fixes SonarQube security-hotspot findings across the CI workflows and Docker images: unpinned installs and source builds that let a dependency's `setup.py`/build backend run arbitrary code.

- **CI workflows** (`publish_docs.yml`, `pypi-publish.yml`): pin mkdocs deps to poetry.lock versions, restrict to prebuilt wheels, and replace the `curl | python3` Poetry installer with a pinned, TLS/hash-verified `pip install poetry==2.4.1`.
- **App/MCP images** (`Docker/{dev,prod}`, `Docker/mcp/{dev,prod}`): pin every dep with `--only-binary=:all:`, install the locally built wheel via a derived-version `--find-links` pin, and add an `OPTICS_FRAMEWORK_VERSION` build arg for the prod images that pulled from PyPI unpinned.
- **Mock OCR images** (`tools/mock_remote_ocr/scalable/Dockerfile{,.cuda}`): copy the full source so the wheelhouse builds a real wheel, drop the `|| true` fallback, and install wheel-only from that wheelhouse.

Also documents a local SonarQube scan (via `sonarsource/sonar-scanner-cli`) in the developer guide for catching these findings before pushing.

## Notes

- `--only-binary=:all:` is applied wherever a *named* third-party package is installed; local source (`pip wheel .`) still builds, since that's the repo's own unpublished code, not a third-party build.
- `jsmin`/`csscompressor` are sdist-only on PyPI, so they're carved out with `--no-binary`.
- `poetry install --sync` has no wheel-only equivalent (it builds via each package's PEP 517 backend); accepted as residual risk on the ephemeral release runner.

## Validation

- Version pins checked against a live `pip`/`uv --only-binary=:all:` resolution on `python:3.12-slim`.
- `Docker/mcp/dev/Dockerfile` built end-to-end; `pre-commit` clean on all changed files.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -24,7 +24,10 @@ jobs:
           path: .cache
           restore-keys: |
             mkdocs-material-
-      - run: pip install mkdocs-material mkdocstrings[python] mkdocs-minify-plugin
+      # jsmin/csscompressor are sdist-only on PyPI, hence --no-binary.
+      - run: |
+          pip install --no-cache-dir --only-binary=:all: --no-binary=jsmin,csscompressor \
+            mkdocs-material==9.7.6 "mkdocstrings[python]==1.0.4" mkdocs-minify-plugin==0.8.0
       - run: mkdocs build
 
   deploy:
@@ -51,5 +54,7 @@ jobs:
           path: .cache
           restore-keys: |
             mkdocs-material-
-      - run: pip install mkdocs-material mkdocstrings[python] mkdocs-minify-plugin
+      - run: |
+          pip install --no-cache-dir --only-binary=:all: --no-binary=jsmin,csscompressor \
+            mkdocs-material==9.7.6 "mkdocstrings[python]==1.0.4" mkdocs-minify-plugin==0.8.0
       - run: mkdocs gh-deploy --force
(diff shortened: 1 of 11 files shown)


malto101:
few things to point,
1. going to be nagging, but commits can be merged, as its a new addition, you dont need split it this intensely, only time that makes sense is when prior main addition got merged
2. instead of local sonar shell, why not move it to sonar docker container? it will be just one cmd, we can add it as part of dev docs
```
docker run \
    --rm \
    -v "${YOUR_REPO}:/usr/src" \
    --network="host" \
    -e SONAR_HOST_URL="${SONARQUBE_URL}" \
    -e SONAR_SCANNER_OPTS="-Dsonar.projectKey=${YOUR_PROJECT_KEY}" \
    -e SONAR_TOKEN="${YOUR_TOKEN}" \
    sonarsource/sonar-scanner-cli
```
3. There's too many unnecessary explanations/comments in CI files

Rest all looks solid

chinmayajha:
> Rest all looks solid

Feedback taken, will make the changes soon @malto101

chinmayajha:
@malto101 I have squashed the commits, documented sonar-scanner-cli container in the developer guide, either trimmed or removed the comments. 

Also found a dependency issue, have fixed it too. Please review once you can. Thanks!

chinmayajha:
@malto101 please let me know if any further changes are needed

thakur-patel (approved):

PR #410, opened 2026-07-28
fix(appium): honor index in AppiumFindElement Text branch

kun-codes:
## Summary

The `Text` (accessibility-id) branch of `AppiumFindElement.locate()` ignored the `index` argument, so `press_element` / `press_element_with_index` could never disambiguate among multiple elements sharing the same accessibility id — it always returned the first match.

## Root cause

```python
found_element = driver.find_element(AppiumBy.ACCESSIBILITY_ID, locator)  # singular, index dropped
return found_element
```

`find_element` (singular) returns only the first match, and `index` was never consulted. So `index="2"` behaved identically to `index="0"`. The sibling `Class` branch already did this correctly with `find_elements` + `index`; the `Text` branch was simply inconsistent.

### DIFF
--- a/optics_framework/engines/elementsources/appium_find_element.py
+++ b/optics_framework/engines/elementsources/appium_find_element.py
@@ -113,8 +113,16 @@ class AppiumFindElement(ElementSourceInterface):
         elif element_type == 'Text':
             locator = element[len("text="):] if element.lower().startswith("text=") else element
             try:
-                found_element = driver.find_element(AppiumBy.ACCESSIBILITY_ID, locator)
-                return found_element
+                found_elements = driver.find_elements(AppiumBy.ACCESSIBILITY_ID, locator)
+                index_to_use = 0 if index is None else index
+                if not found_elements or index_to_use < 0 or index_to_use >= len(found_elements):
+                    raise OpticsError(
+                        Code.E0201,
+                        message=f"Element of type {element_type} and index {index_to_use} not found using: {element}",
+                    )
+                return found_elements[index_to_use]
+            except OpticsError:
+                raise
             except NoSuchElementException as e:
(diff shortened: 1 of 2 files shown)


optics_framework/engines/elementsources/appium_find_element.py
```diff
@@ -113,8 +113,16 @@ def locate(self, element: str, index: Optional[int] = None) -> Any:
...
             try:
-                found_element = driver.find_element(AppiumBy.ACCESSIBILITY_ID, locator)
-                return found_element
+                found_elements = driver.find_elements(AppiumBy.ACCESSIBILITY_ID, locator)
+                index_to_use = 0 if index is None else index
+                if not found_elements or index_to_use < 0 or index_to_use >= len(found_elements):
+                    raise OpticsError(
+                        Code.E0201,
+                        message=f"Element of type {element_type} and index {index_to_use} not found using: {element}",
+                    )
```
chinmayajha, line 122: good work

2 follow-ups -
- Could you add a regression test to `tests/units/engines/elementsources/test_appium_find_element.py`? There is currently no coverage of `locate()` index behavior for either the Text or Class branch — a test asserting `index=1` selects the second accessibility-id match (and out-of-range → `E0201`) would lock this in.
- Minor/optional: `except NoSuchElementException` is now effectively unreachable here, since `find_elements` returns `[]` instead of raising. Fine to keep for parity with the `Class` branch — just noting it.

chinmayajha (approved):
Seems good to me

chinmayajha (approved):
Good fix, once tests are added, can be merged

optics_framework/engines/elementsources/appium_find_element.py
kun-codes: Fixed. Have a look at d4d3899064c3dd298552bc842dd274294b8f7850

PR #411, opened 2026-07-28
fix(actions): apply offset_x/offset_y to element-resolved presses

kun-codes:
## Summary

`press_element`'s `offset_x` / `offset_y` were silently ignored whenever the element resolved to a WebElement (accessibility / XPath match). The offset only ever took effect for coordinate-resolved matches (OCR / image), so an offset press against an accessibility- or XPath-located element just clicked the element centre.

## Root cause

```python
if isinstance(located, tuple):
    x, y = located
    self.driver.press_coordinates(x + int(offset_x), y + int(offset_y), event_name)  # offset applied
else:
    self.driver.press_element(located, int(repeat), event_name)                      # WebElement → offset dropped
```

`located` is a `(x, y)` tuple only for coordinate strategies; for accessibility/XPath it is a WebElement, and that branch never consulted `offset_x` / `offset_y`. So the offset reached the tap only for OCR/image matches and was discarded for every element-resolved match.

### DIFF
--- a/optics_framework/api/action_keyword.py
+++ b/optics_framework/api/action_keyword.py
@@ -464,6 +464,23 @@ class ActionKeyword:
         if screenshot_np is not None:
             utils.save_screenshot(screenshot_np, name, output_dir=self.execution_dir)
 
+    @staticmethod
+    def _element_centre(element: Any) -> Optional[Tuple[int, int]]:
+        """Centre (x, y) of a located WebElement in the driver's coordinate space.
+
+        Reads location/size (available on both Appium and Selenium WebElements);
+        returns None if either is unavailable, so callers can fall back.
+        """
+        try:
+            loc = element.location
+            size = element.size
+            return (
+                int(loc["x"] + size["width"] / 2),
+                int(loc["y"] + size["height"] / 2),
+            )
+        except (AttributeError, KeyError, TypeError):
+            return None
+
     # Click actions
     @with_self_healing
     def press_element(
@@ -485,12 +502,28 @@ class ActionKeyword:
         :param aoi_width: Width percentage of Area of Interest (0-100). Default: 100.
         :param aoi_height: Height percentage of Area of Interest (0-100). Default: 100.
         """
+        offset_x_i, offset_y_i = int(offset_x), int(offset_y)
         if isinstance(located, tuple):
             x, y = located
             internal_logger.info(
-                f"Pressing at coordinates ({x + int(offset_x)}, {y + int(offset_y)}) with offset ({offset_x}, {offset_y})")
+                f"Pressing at coordinates ({x + offset_x_i}, {y + offset_y_i}) with offset ({offset_x_i}, {offset_y_i})")
             self.driver.press_coordinates(
-                x + int(offset_x), y + int(offset_y), event_name)
+                x + offset_x_i, y + offset_y_i, event_name)
+        elif offset_x_i or offset_y_i:
+            # located is a WebElement (accessibility/XPath). Clicking it directly would
+            # ignore offset_x/offset_y, so derive its centre and press by coordinates
+            # with the offset applied.
+            centre = self._element_centre(located)
+            if centre is not None:
+                cx, cy = centre
+                internal_logger.info(
+                    f"Pressing at coordinates ({cx + offset_x_i}, {cy + offset_y_i}) with offset ({offset_x_i}, {offset_y_i})")
+                for _ in range(int(repeat)):
+                    self.driver.press_coordinates(cx + offset_x_i, cy + offset_y_i, event_name)
+            else:
+                internal_logger.warning(
+                    f"Could not determine element centre for '{element}'; pressing element directly (offset ignored).")
+                self.driver.press_element(located, int(repeat), event_name)
         else:
             internal_logger.info(f"Pressing element '{element}'")
             self.driver.press_element(located, int(repeat), event_name)
--- a/optics_framework/api/action_keyword.py
+++ b/optics_framework/api/action_keyword.py
@@ -466,13 +466,13 @@ class ActionKeyword:
 
     @staticmethod
     def _element_centre(element: Any) -> Optional[Tuple[int, int]]:
-        """Centre (x, y) of a located WebElement in the driver's coordinate space.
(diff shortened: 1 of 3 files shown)


optics_framework/api/action_keyword.py
```diff
@@ -485,12 +502,28 @@ def press_element(
         :param aoi_width: Width percentage of Area of Interest (0-100). Default: 100.
         :param aoi_height: Height percentage of Area of Interest (0-100). Default: 100.
         """
+        offset_x_i, offset_y_i = int(offset_x), int(offset_y)
```
chinmayajha, line 511: Seems mergeable once the following are done

- Going through `press_coordinates` skips the scroll-into-view you get from `element.click()`, so an offset press on a partly off-screen element could land in the wrong spot. Worth a comment at least, or handling it.
- Can you add a test with a mock WebElement (`.location`/`.size`) asserting `press_coordinates` gets centre + offset?

chinmayajha (changes requested):
Good catch and fix 

Holding on two things (mentioned below

optics_framework/api/action_keyword.py
kun-codes: Fixed in 0cf4c8f and 247e853

optics_framework/api/action_keyword.py
```diff
@@ -464,6 +464,23 @@ def _save_screenshot_if_available(self, screenshot_np: Any, name: str) -> None:
...
+        """
+        try:
+            loc = getattr(element, "location_once_scrolled_into_view", None) or element.location
+            size = element.size
+            return (
+                int(loc["x"] + size["width"] / 2),
+                int(loc["y"] + size["height"] / 2),
+            )
+        except (AttributeError, KeyError, TypeError):
+            return None
```
chinmayajha: This makes offset presses fail outright on native mobile. `location_once_scrolled_into_view` is JS-backed (selenium executes `arguments[0].scrollIntoView(true); return arguments[0].getBoundingClientRect()`), so under Appium native it raises `WebDriverException` — not in the caught tuple, so it escapes `_element_centre`, escapes `press_element`, and `_try_results_until_success` turns it into `X0201: All strategies failed for '//zone' in 'press_element'`. Repro'd locally: on main the offset was ignored and the tap landed on the centre; now the step fails.

The MagicMock tests can't catch this — every attribute on a MagicMock succeeds.

Catch broadly and fall through to `element_source.get_bbox_for_element` (`elementsource_interface.py:83`), which already handles Appium `bounds`/`rect` and Playwright `bounding_box()`. That also makes offsets work on Playwright, where a `Locator` has no `.location`/`.size` so the offset is still dropped today. It also drops the `.location` fallback, which on Selenium web returns document coords while `press_coordinates` uses `elementFromPoint` (viewport) — wrong tap on a scrolled page.

```suggestion
    def _element_centre(self, element: Any) -> Optional[Tuple[int, int]]:
        """Centre (x, y) of a located element, or None if geometry is unavailable.

        Prefers ``location_once_scrolled_into_view``: on web it scrolls the element
        on-screen and returns viewport coords, which is what ``press_coordinates``
        expects. It is JS-backed, so it raises in a native Appium context - fall back
        to the source-agnostic bbox accessor there.
        """
        try:
            loc = element.location_once_scrolled_into_view
            size = element.size
            return (
                int(loc["x"] + size["width"] / 2),
                int(loc["y"] + size["height"] / 2),
            )
        except Exception as exc:  # noqa: BLE001 - JS-backed, unsupported on native
            internal_logger.debug(f"Scrolled-into-view geometry unavailable: {exc}")
        bbox = self.element_source.get_bbox_for_element(element)
        if bbox is None:
            return None
        (x1, y1), (x2, y2) = bbox
        return ((x1 + x2) // 2, (y1 + y2) // 2)
```

optics_framework/api/action_keyword.py
```diff
@@ -485,12 +502,27 @@ def press_element(
         :param aoi_width: Width percentage of Area of Interest (0-100). Default: 100.
         :param aoi_height: Height percentage of Area of Interest (0-100). Default: 100.
         """
+        offset_x_i, offset_y_i = int(offset_x), int(offset_y)
         if isinstance(located, tuple):
             x, y = located
             internal_logger.info(
-                f"Pressing at coordinates ({x + int(offset_x)}, {y + int(offset_y)}) with offset ({offset_x}, {offset_y})")
+                f"Pressing at coordinates ({x + offset_x_i}, {y + offset_y_i}) with offset ({offset_x_i}, {offset_y_i})")
             self.driver.press_coordinates(
```
chinmayajha: Coordinate-located presses still ignore `repeat` (one `press_coordinates` call regardless), while the new branch loops. Pre-existing, but now the same keyword honours `repeat` in two of three branches — fold it in here or file a follow-up.

chinmayajha (changes requested):
thanks for the tests coverage

The scroll-into-view fix regresses native mobile: `location_once_scrolled_into_view` is JS-backed and raises under Appium native, and that exception isn't caught, so an offset press now fails as `X0201` instead of tapping the centre. Details inline.

Also update the `press_element` example in `docs/architecture/strategies.md` (~line 750) — it still shows the old two-branch body.

optics_framework/api/action_keyword.py
kun-codes: fixed in f85abe7cc8ce46c7af2bfc4e8ffd3bc39a25f3af

optics_framework/api/action_keyword.py
kun-codes: fixed in 4a5f352ab34a824222a182394daab81b35ccd0aa

kun-codes:
> thanks for the tests coverage
> 
> The scroll-into-view fix regresses native mobile: `location_once_scrolled_into_view` is JS-backed and raises under Appium native, and that exception isn't caught, so an offset press now fails as `X0201` instead of tapping the centre. Details inline.
> 
> Also update the `press_element` example in `docs/architecture/strategies.md` (~line 750) — it still shows the old two-branch body.

fixed in f85abe7cc8ce46c7af2bfc4e8ffd3bc39a25f3af and docs is updated in 8376cfd9df3ac52a3c8b70b2b98018527343a294

PR #412, opened 2026-07-28
fix(strategies): enforce AOI for accessibility/XPath element matches

kun-codes:
## Summary

`press_element`'s `aoi_x` / `aoi_y` / `aoi_width` / `aoi_height` were silently ignored whenever the element resolved via an accessibility or XPath strategy. The AOI only ever constrained the OCR / image strategies, so an AOI-scoped press against an element-source match could resolve and press an element outside the requested region.

## Root cause

In `StrategyManager._try_strategy_locate`, the AOI is only applied to strategies that implement `locate_with_aoi`:

```python
locate_with_aoi = getattr(strategy, "locate_with_aoi", None)
if use_aoi and locate_with_aoi is not None:
    result = locate_with_aoi(element, aoi_x, aoi_y, aoi_width, aoi_height, index=index)
else:
    result = strategy.locate(element, index=index)   # AOI dropped
```

Only `TextDetectionStrategy` (OCR) and `ImageDetectionStrategy` define `locate_with_aoi`. `TextElementStrategy` (accessibility) and `XPathStrategy` do not, and they run first in priority order, so their match was returned with the AOI discarded — the press could land anywhere the element source found the target, regardless of the AOI.

### DIFF
--- a/optics_framework/common/strategies.py
+++ b/optics_framework/common/strategies.py
@@ -604,14 +604,11 @@ class StrategyManager:
     ) -> bool:
         """True if a located WebElement's centre lies inside the AOI rectangle.
 
-        AOI is applied natively by strategies implementing ``locate_with_aoi`` (OCR /
-        image). Accessibility / XPath strategies return a WebElement and would otherwise
-        ignore the AOI, so this filters them by bounding box. The element centre is
-        compared as a percentage of the driver window against the AOI percentages
-        (percentages are resolution-independent, so no screenshot scaling is needed).
+        Filters accessibility/XPath matches (which return a WebElement and ignore the
+        AOI, unlike ``locate_with_aoi`` OCR/image strategies), comparing the centre as a
+        window percentage against the AOI. Fails open when bbox/window is unknown.
 
-        Fails open (returns True) when the bbox or window size can't be determined, so
-        behaviour is unchanged when the check can't run.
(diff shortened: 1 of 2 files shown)


optics_framework/common/strategies.py
```diff
@@ -627,6 +666,12 @@ def _try_strategy_locate(
                 value, annotated_frame = result[0], result[1]
             else:
                 value, annotated_frame = result, None
+            if use_aoi and locate_with_aoi is None and not isinstance(value, tuple):
```
chinmayajha, line 666: Add a test: stub strategy with one bbox inside the AOI, one outside — assert the outside one is skipped and falls through. Nothing covers this path.

Minor: centre-based check can disagree with the OCR crop for an element on the AOI edge — worth a docstring line.

chinmayajha (approved):
Valid fix 
AOI was ignored for accessibility/XPath matches. Only thing is to adda test for the in/out-of-AOI fall-through

malto101:
Hello @kun-codes 
thanks for the fix
AOI is good feature, but I feel the UX is bad
currently It is taking starting x, starting y, width, height 
which isnt something a user who just has a screen it front of him will make good of

could you think of multiple ways, we can improve the UX in this case as a separate PR/issue?

malto101 (approved):
LGTM
just need the coverage as written by chimney

Issue #417, opened 2026-07-29
Optimize: skip screenshot capture in get_interactive_elements when pixel-space bounds aren't needed


chinmayajha:
### Context

Follow-up from the interactive-element bounds pixel-scaling fix (PR to link below).

`Verifier.get_interactive_elements` now captures a screenshot on Appium sources so element `bounds` can be scaled from the driver's window coordinate space into the screenshot's pixel space (necessary for correct overlays on Retina iOS). See `Verifier._bounds_need_screenshot` / `_collect_interactive_elements` in `optics_framework/api/verifier.py`.

### Problem

Previously `get_interactive_elements` was page-source only. The added capture is a real device round-trip and now runs on **every** Appium call, including callers that never read `bounds`:

- the CSV/YAML runner keyword
- `optics live`
- dropdown-option enumeration (`_safe_get_interactive_elements`, `optics_framework/api/action_keyword.py:645`)

For those, the capture + scale is wasted work.

### Proposal

Add an opt-out so callers that only need `text`/`xpath` can skip the capture and scaling, e.g. a `scale_bounds: bool = True` parameter on `get_interactive_elements` (default preserves the corrected behavior), and pass `False` from the internal text/xpath-only call sites.

### Notes / edge cases also worth tracking

- Not yet verified end-to-end on a real iOS device (XCUITest points vs screenshot pixels); math is unit-tested only.
- `_resolve_source_instance` falls back to `instances[0]` when `current_instance` is cleared — with multiple enabled Appium element sources the window size could be read from the wrong device.

PR #418, opened 2026-07-29
fix: return interactive-element bounds in screenshot pixel space

chinmayajha:
## Problem

`get_interactive_elements`, `get_screen_elements`, and the `optics serve` workspace stream return element `bounds` sourced from the page source — the driver's **window coordinate space** (points on iOS, pixels on Android) — alongside a screenshot in **pixel space**. On Retina iOS these spaces differ by the device scale factor, so a client overlaying the bounds on the screenshot mis-places every box. On Android the two coincide, so it's a no-op today.

## Approach

Scale the bounds into the screenshot's pixel space inside the framework, so API/workspace consumers can draw them directly with no scaling of their own.

Main already had a parallel scaler for the **annotation** path (`scale_bboxes_for_screenshot`, tuple-shaped, drawn onto saved screenshots). Rather than add a second copy of the same arithmetic for the dict-shaped API bounds, this PR reconciles both behind **one gated ratio helper** so they can't drift:

- `utils._pixel_scale_for_source` — resolves an `InstanceFallback` wrapper to its concrete source, gates to **Appium** (whose window and element coordinates share one space; web sources report them in unrelated spaces so the ratio is meaningless), reads the window size, returns per-axis `(scale_x, scale_y)` or `None`.
- `scale_bboxes_for_screenshot` (tuples) and the new `scale_interactive_element_bounds` (dicts) are thin adapters over it.
- The shared gate also corrects the annotation path, which previously scaled Selenium bboxes with no gate.
- `Verifier` derives the returned image and the bounds from a **single** capture; `get_interactive_elements` only captures on Appium (no-op elsewhere).

## Commits

1. `refactor(utils)` — unify scaling behind the gated ratio helper.
2. `fix(verifier)` — pixel-space bounds from a shared capture.
3. `fix(serve)` — workspace stream aligns bounds with the returned frame.

## Tests

`tests/units/common/test_bbox_scaling.py` (annotation path, incl. non-Appium regression guard) and `test_interactive_bounds_scaling.py` (API path + cross-path consistency). Full unit suite green (610 passed).

## Follow-ups

- #417 — extra screenshot capture on `get_interactive_elements` for callers that only need text/xpath; also tracks real-device verification and the multi-source `current_instance` edge case.

## Notes for reviewers

- Behavior change for HiDPI Selenium annotations (now un-scaled) — a no-op when `devicePixelRatio == 1`.

### DIFF
--- a/optics_framework/common/utils.py
+++ b/optics_framework/common/utils.py
@@ -942,31 +942,61 @@ def _window_size_from_source(element_source: Any) -> Optional[Tuple[int, int]]:
     return None
 
 
-def _scale_bbox(
+def _pixel_scale_for_source(
+    element_source: Any,
+    screenshot: Optional[np.ndarray],
+) -> Optional[Tuple[float, float]]:
+    """
+    Compute the per-axis scale that maps the driver's window coordinate space onto
+    the screenshot's pixel space, or ``None`` when scaling should not be applied.
+
+    * Gated to **Appium** sources. Their page-source/WebElement coordinates and the
+      driver's window size share one coordinate system (points on iOS, pixels on
+      Android), so ``screenshot / window`` is a meaningful ratio. Web sources
+      (Selenium/Playwright) report element coordinates (CSS pixels, scroll-relative)
+      and window size in unrelated spaces, so the ratio is not correct there and we
+      return ``None`` to leave their coordinates untouched.
+    * Best-effort and non-failing: returns ``None`` when no screenshot is available,
+      the window size can't be determined, or its shape can't be read — callers then
+      leave coordinates unscaled rather than raising.
+
+    :param element_source: Element source (possibly an ``InstanceFallback`` wrapper).
+    :param screenshot: The captured frame whose pixel space is the target.
+    :return: ``(scale_x, scale_y)``, or ``None`` when scaling must not be applied.
+    """
+    if screenshot is None:
+        return None
+    source = getattr(element_source, "active_instance", None) or element_source
+    if getattr(source, "REQUIRED_DRIVER_TYPE", None) != "appium":
+        return None
+    window_size = _window_size_from_source(source)
+    if window_size is None:
+        return None
+    window_width, window_height = window_size
+    try:
+        screenshot_height, screenshot_width = screenshot.shape[:2]
+    except (AttributeError, TypeError):
+        return None
+    if window_width <= 0 or window_height <= 0:
+        return None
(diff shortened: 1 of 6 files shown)


optics_framework/common/utils.py
```diff
@@ -942,31 +961,65 @@ def _window_size_from_source(element_source: Any) -> Optional[Tuple[int, int]]:
...
+    element_source: Any,
+    screenshot: Optional[np.ndarray],
+) -> Optional[Tuple[float, float]]:
+    """
+    Compute the per-axis scale that maps the driver's window coordinate space onto
+    the screenshot's pixel space, or ``None`` when scaling should not be applied.
+
+    This is the single decision point shared by every bounds/bbox scaler
+    (``scale_bboxes_for_screenshot`` for annotation, ``scale_interactive_element_bounds``
+    for API responses), so the gating rules live in exactly one place:
```
thakur-patel: so the gating rules live in exactly one place - this is relevant to the PR, not to the entire repo
we can delete that

optics_framework/common/utils.py
```diff
@@ -942,31 +961,65 @@ def _window_size_from_source(element_source: Any) -> Optional[Tuple[int, int]]:
...
+        return None
+    source = _resolve_source_instance(element_source)
+    if getattr(source, "REQUIRED_DRIVER_TYPE", None) != "appium":
+        return None
+    window_size = _window_size_from_source(source)
+    if window_size is None:
+        return None
+    win_w, win_h = window_size
+    try:
+        sh_h, sh_w = screenshot.shape[:2]
```
thakur-patel: let's avoid having names this brief. a bit more descriptive would help.

optics_framework/common/utils.py
```diff
@@ -910,6 +910,25 @@ def bbox_from_webelement_like(obj: Any) -> Optional[Tuple[Tuple[int, int], Tuple
     return None
 
 
+def _resolve_source_instance(element_source: Any) -> Any:
```
thakur-patel: this feels like a getter. do we need a separate function in util() for this or can we get this as an attribute of a certain object?

optics_framework/common/utils.py
chinmayajha: Fixed

optics_framework/common/utils.py
chinmayajha: Fixed

thakur-patel (approved):
LGTM

PR #419, opened 2026-07-30
feat(setup): allow pinning an engine version in `optics setup --install`

### LINKED ISSUE #355: bug: deps installed via `optics setup` do not have pinned versions
**Describe the bug**
dependencies installed via `optics setup` do not have pinned versions. The latest version of dependencies get installed all the time which can be a problem for testing in ci/cd

**To Reproduce**
1. Run commands like `optics setup --install "EasyOCR" "Appium"`
2. Latest version of `easyocr` and `appium` get installed

**Expected behavior**
A pinned version gets installed. If required, user should be able to override the version of a dependency with their required version

chinmayajha:
## Summary

Closes #355.

`optics setup --install` already installs engines through version-bounded pyproject extras (pinned to your installed Optics version), so the "always latest, unbounded" part of #355 is handled. What was still missing is the issue's second ask: **letting a user pin a specific dependency version** — needed for reproducible CI/CD.

Any `--install` token may now carry a PEP 508 specifier:

```bash
optics setup --install appium==4.2.0 "easyocr>=1.7,<2.0"
```

## How it works

- `resolve_engines` splits the specifier off each token into a new `InstallRequest(engine, version)`.
- `install_extras` keeps the extras spec (`optics-framework[appium,easyocr]==<version>`) for transitive deps and, for any request with a version, appends the engine's **primary package** as a concrete pin (`appium-python-client==4.2.0`). pip intersects that with the extra's own range, so an out-of-range pin fails loudly rather than silently resolving elsewhere.
- A specifier on a bundle (`all==1.0`) is rejected — one version can't span many packages.
- The TUI picker (no version field) installs at the extra's default bound, unchanged.

## Tests

Extended `tests/units/helpers/test_setup.py`: specifier parsing (`_split_token`), version threaded onto `InstallRequest`, bundle+version rejection, explicit-version-wins dedupe, and the pinned-package install command. All 43 pass; `pre-commit` clean.

## Note (release timing)

The extras themselves land with the unreleased `1.9.0-beta.1`; the latest PyPI release (`1.8.6`) predates them, so `optics setup` only resolves extras once a build carrying them is published. This PR is about the override capability and is orthogonal to that release.

### DIFF
--- a/optics_framework/helper/setup.py
+++ b/optics_framework/helper/setup.py
@@ -5,10 +5,17 @@ from typing import Dict, List, Optional
 import sys
 from textual.app import App, ComposeResult
 from textual.widgets import Checkbox, Button, Header, Footer, Static
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
 from pydantic import BaseModel
 
 
 DISTRIBUTION_NAME = "optics-framework"
+
+
+class SetupError(Exception):
+    """A user-facing engine-setup error (bad version specifier, version
+    conflict). Raised by `resolve_engines` and surfaced by the CLI as a clean
+    message instead of a Python traceback."""
 
 
 class EngineBackend(BaseModel):
@@ -32,7 +39,7 @@ class EngineCategory(BaseModel):
 class InstallRequest(BaseModel):
     """One engine to install, optionally with a user-supplied version specifier.
 
-    ``version`` is the raw PEP 508 tail (e.g. ``"==4.2.0"``, ``">=4.0,<4.3"``)
+    ``version`` is the raw PEP 508 tail (e.g. ``"==5.0.0"``, ``">=5.0,<5.4"``)
     parsed off the CLI token; ``None`` means "let the extra's own bound decide"."""
     engine: EngineBackend
     version: Optional[str] = None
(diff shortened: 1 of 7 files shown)


optics_framework/helper/setup.py
```diff
@@ -97,32 +123,60 @@ def _alias_index() -> Dict[str, EngineBackend]:
     return index
 
 
-def resolve_engines(tokens: List[str]) -> tuple[List[EngineBackend], List[str]]:
-    """Resolve user-supplied tokens to EngineBackends. Returns (resolved, invalid).
+def _add_request(resolved: List[InstallRequest], engine: EngineBackend, version: Optional[str]) -> None:
```
kun-codes, line 146: I tried the command `optics setup --install appium==4.2.0 appium==5.0.0` and I expected appium 5.0.0 to be installed since it is located at the last but 4.2.0 was installed

```
pip show appium-python-client
Name: Appium-Python-Client
Version: 4.2.0
Summary: Python client for Appium
Home-page: http://appium.io/
Author: Isaac Murchie
Author-email: isaac@saucelabs.com
License: Apache 2.0
Location: /Users/user/Library/Caches/pypoetry/virtualenvs/optics-framework--OROUlrp-py3.14/lib/python3.14/site-packages
Requires: selenium
Required-by:
```

optics_framework/helper/setup.py
```diff
@@ -86,6 +96,22 @@ def _norm(token: str) -> str:
     return token.strip().lower().replace(" ", "_").replace("-", "_")
 
 
+# First char that starts a PEP 508 version specifier ( ==, >=, ~=, !=, <, > ).
+_SPEC_START = re.compile(r"[=<>!~]")
+
+
+def _split_token(token: str) -> tuple[str, Optional[str]]:
```
kun-codes, line 110:
```
optics setup --install appium=4.2.0
Installation failed: Command '['/Users/inno/Library/Caches/pypoetry/virtualenvs/optics-framework--OROUlrp-py3.14/bin/python', '-m', 'pip', 'install', 'optics-framework[appium]==1.9.0b1', 'appium-python-client=4.2.0']' returned non-zero exit status 1.
[notice] A new release of pip is available: 26.1.2 -> 26.2
[notice] To update, run: pip install --upgrade pip
ERROR: Invalid requirement: 'appium-python-client=4.2.0': Expected semicolon (after name with no version specifier) or end
    appium-python-client=4.2.0
                        ^
Hint: = is not a valid operator. Did you mean == ?
```

the code of optics considered it as valid code when it isn't so it is passed to python where it fails. It is not valid since `=` is not considered valid

kun-codes (changes requested):

optics_framework/helper/setup.py
chinmayajha: Fixed — two different versions for one engine now raise a conflict (like pip's `pkg==a pkg==b`) instead of silently keeping the first.

optics_framework/helper/setup.py
chinmayajha: Fixed — specifiers are validated with `packaging.SpecifierSet` up front, so `appium=4.2.0` (single `=`) now errors cleanly from optics instead of reaching pip.

kun-codes (approved):

PR #425, opened 2026-08-03
fix(docker): bump appium-python-client pin to 5.3.1

chinmayajha:
## Summary
- All four Dockerfiles pinned `appium-python-client==4.5.1`, which is no longer on PyPI (only 5.2.0+ remain), breaking every image build
- Bumped to `5.3.1`, satisfying `pyproject.toml`'s `>=5.0,<6.0` constraint

## Test plan
- [x] `docker build` succeeds for each of `Docker/dev`, `Docker/prod`, `Docker/mcp/dev`, `Docker/mcp/prod`

### DIFF
--- a/Docker/dev/Dockerfile
+++ b/Docker/dev/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 # Install common Python packages and optics-framework wheel and vision backend
 RUN pip install --no-cache-dir --only-binary=:all: \
-   appium-python-client==4.5.1 playwright==1.61.0 \
+   appium-python-client==5.3.1 playwright==1.61.0 \
    && playwright install-deps chromium firefox \
    && WHEEL=$(if [ -n "$WHL_FILE" ]; then echo "/app/${WHL_FILE}"; else ls /app/*.whl 2>/dev/null | head -1; fi) \
    && if [ -z "$WHEEL" ] || [ ! -f "$WHEEL" ]; then \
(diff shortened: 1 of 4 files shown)


thakur-patel:
https://github.com/mozarkai/optics-framework/pull/408/changes#diff-a78217d24faa7fc0eb1fe5aa0b994771e3d6ec9d6dc54e60498dae188c52d2d0 

The pinning was added in this PR. Can we have a single source of truth for version pinning?

chinmayajha:
Yes, planning to land that next @thakur-patel

chinmayajha:
@thakur-patel PR [#408](https://github.com/mozarkai/optics-framework/pull/408)'s branch was cut from a commit dated 2026-07-23, when pyproject.toml constrained appium-python-client to >=3.1,<5.0. 4.5.1 was poetry's resolved version under that constraint at the time, so it was a correct, consistent pin when the branch was created.

Why it broke: Three days later, on 2026-07-28, a separate commit — [8ac4707](https://github.com/mozarkai/optics-framework/commit/8ac4707) "build(deps): require appium-python-client >=5.0" because the code needed feature introduced post Appium 5.0.0. So we raised the floor in pyproject.toml to >=5.0,<6.0.

So the newly-added --only-binary=:all: constraint from PR [#408](https://github.com/mozarkai/optics-framework/pull/408) was never going to work with 4.5.1 regardless of the version-floor bump

Why nothing caught this: none of the .github/workflows/* build these Docker images — there's no CI step that exercises docker build on a PR.

thakur-patel (approved):

PR #433, opened 2026-08-04
fix(docker): install CPU-only torch before easyocr

thakur-patel:
## Summary
- `easyocr` transitively depends on `torch`, and PyPI's default Linux `torch` wheel bundles the full CUDA toolkit as separate `nvidia-*` packages (cuBLAS, cuSOLVER, cuDNN, NCCL, etc). These containers only run OCR on CPU, so that's several unused GB — large enough that it filled a build host's disk mid-image-export during a real build attempt.
- Installs the CPU-only `torch`/`torchvision` wheel from `download.pytorch.org/whl/cpu` before `easyocr` in `Docker/prod/Dockerfile`, `Docker/mcp/prod/Dockerfile`, and `Docker/dev/Dockerfile`, so pip/uv never resolves the CUDA variant.
- Documents a branch-naming convention in `CLAUDE.md` (`fix/`, `feat/`, `refactor/`, etc.) mirroring the Conventional Commit types already enforced by the commit-msg hook.
- Documents an image-tagging convention in `Docker/deployment.md`: tag images `<optics_version>-<vision_backend>` (e.g. `1.9.1-easyocr`) across all four build paths, so the tag itself states what's baked into the image.

## Why
A production image build from a disk-constrained host (Cloud Shell, 95G root) failed with `no space left on device` during image export. Root cause traced to the CUDA-enabled `torch` wheel pulled in transitively by `easyocr` — none of the CUDA libraries are used since these images run inference on CPU only.

## Test plan
- [x] Rebuilt `Docker/prod/Dockerfile` locally (`VISION_BACKEND=easyocr`, the default) and confirmed the image builds without pulling any `nvidia-*` CUDA packages (`torch==2.13.0+cpu`, `torchvision==0.28.0+cpu`), and that `easyocr.Reader(['en'])` still initializes fine on CPU torch.
- [ ] Repeat the same check for `Docker/mcp/prod/Dockerfile` and `Docker/dev/Dockerfile`.


malto101 (approved):
LGTM
could you fix those warnings raised by sonarqube

malto101:
seems to have diverge, @thakur-patel could you rebase from main tip

PR #434, opened 2026-08-06
fix/select highest confidence match in easyocr and remoteocr

kun-codes:
This partially addresses a bug which I found while writing tests for running on emulated devices. I found that in case of multiple matches in the image, the first detected match would be selected which may or may not have the highest confidence. This pr sorts the matches with highest confidence and will select the match with the highest confidence

I would be creating issues for Google Vision and Pytesseract which I will link here

Edit: Issues #435 #436 and #437 are created


malto101:
Hello @kun-codes 
could you explain how exactly you solved this.
before we review it

PR #438, opened 2026-08-06
fix/letterboxed window bounds scaling

kun-codes:
This pr fixes the bug which returns wrong bounding boxes for ios when the app is letterboxed


malto101 (approved):
Lgtm

thakur-patel (approved):

PR #444, opened 2026-08-09
fix(setup): make the engine-installer TUI responsive and legible

chinmayajha:
## What

Bare `optics setup` opens a Textual picker that *looked* broken — the handler chain is wired correctly (checkbox → selection → `install_extras`), but three UX bugs made pressing **Install Selected** feel like a no-op:

1. **UI freeze.** The install ran `subprocess.run(pip install …)` synchronously on Textual's event loop, so the whole TUI hung for the duration (and again for the Playwright browser download).
2. **Invisible feedback.** Success/failure went through `print()`, which is swallowed while Textual owns the alternate screen — the screen looked unchanged.
3. **Unreachable buttons.** The checkbox list wasn't in a scroll container, so on a short terminal (e.g. 80×24) the buttons fell below the fold.

## Fix

- Run the install in a `@work(thread=True)` worker; disable **Install** and show an "Installing…" status while it runs, then report the result in a status line + `notify(...)` and re-enable for retry.
- `install_extras` now returns `(success, message)` instead of only printing. The CLI `--install` path prints that message and **exits non-zero on failure** (it previously exited 0 even when pip failed — bad for scripts/CI).
- Wrap the checkbox list in a `VerticalScroll` so the buttons stay reachable.

## Testing

- New `TestEngineInstallerApp` (Textual `run_test` harness): the list is in a `VerticalScroll` and a status line exists; no-selection notifies without installing; a selected install runs in a worker and shows the result; a failed install shows the error and re-enables Install.
- Updated `install_extras` tests for the `(success, message)` contract; full `test_setup.py` = 53 passed.
- `pre-commit` clean; CLI: failed `--install` now exits 1, `--list` exits 0.

## Context

Reported from a headless probe of the picker; part of the same README/onboarding hardening pass as the companion PRs.

---
**Follow-ups:** #447 (stream long-install progress; report installed extras)

### DIFF
--- a/optics_framework/helper/cli.py
+++ b/optics_framework/helper/cli.py
@@ -384,7 +384,10 @@ class EngineInstaller(Command):
                 print(f"Error: Invalid engine(s): {', '.join(invalid)}")
                 print("Use `optics setup --list` to see available engines")
                 return
-            install_extras(engines)
+            success, message = install_extras(engines)
+            print(message)
+            if not success:
+                sys.exit(1)
         else:
             EngineInstallerApp().run()
 
--- a/optics_framework/helper/setup.py
+++ b/optics_framework/helper/setup.py
@@ -3,7 +3,9 @@ import subprocess  # nosec B404
 from importlib.metadata import PackageNotFoundError, version
 from typing import Dict, List, Optional
 import sys
+from textual import work
 from textual.app import App, ComposeResult
+from textual.containers import VerticalScroll
 from textual.widgets import Checkbox, Button, Header, Footer, Static
 from packaging.specifiers import InvalidSpecifier, SpecifierSet
 from pydantic import BaseModel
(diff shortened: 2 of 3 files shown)


malto101:
@chinmayajha can you share some sample screenshots usage and also add those to docs to where the TUI is present.

chinmayajha:
> @chinmayajha can you share some sample screenshots usage and also add those to docs to where the TUI is present.

@malto101  Not uploading to docs, as it is pretty evident once people run optics setup. I'll add a better hint towards running it though, that I'll add. 

image

image

image

chinmayajha:
@malto101 bump

thakur-patel (approved):

malto101:
@chinmayajha could you add a small gift/photo in docs

malto101 (approved):

malto101:
Damn
PR no 444!

PR #448, opened 2026-08-12
fix(serve): make optics loop non blocking enabling health checks while loop is on

chinmayajha:
## What changed

`optics serve` no longer runs blocking driver/Appium work on its asyncio event-loop thread:

- session construction, including the Appium new-session handshake, runs in a worker thread;
- individual keyword calls run in a worker thread;
- an `asyncio.Lock` serializes keyword execution per Optics session, preserving the existing one-command-at-a-time WebDriver semantics.

## Why

Synchronous Appium I/O could starve the serve loop long enough for `/` and `/health` to time out. In Kubernetes this made a live worker look dead to readiness/liveness probes; in the fleet deployment it could also trigger an unnecessary gateway reattach.

The normal `optics execute` batch and dry-run flows are unaffected: they use `BatchExecutor`/`TestRunner`, not `KeywordExecutor`.

## Validation

- Added a regression test proving serve keyword work runs outside the event-loop thread and concurrent requests to one session remain serialized.
- Extended the session-start endpoint test to prove driver construction is offloaded from the API loop.
- `pytest tests/units -q` (pass; one existing Playwright-dependent skip and two existing expected failures).
- `pre-commit run --files ...` (ruff, bandit, whitespace, EOF, secret scan all pass).

### DIFF
--- a/optics_framework/common/execution.py
+++ b/optics_framework/common/execution.py
@@ -186,7 +186,10 @@ class KeywordExecutor(Executor):
             try:
                 # Deserialize parameters based on method signature
                 deserialized_params = _deserialize_params(method, self.params)
-                result = method(*deserialized_params)
+                async with session.keyword_lock:
+                    result = await asyncio.to_thread(method, *deserialized_params)
+                    heal_owner = getattr(method, "__self__", None)
+                    heal_info = heal_owner._pop_last_heal_info() if hasattr(heal_owner, "_pop_last_heal_info") else None
             except Exception as e:
                 await event_manager.publish_event(Event(
                     entity_type="keyword",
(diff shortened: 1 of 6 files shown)


optics_framework/common/execution.py
```diff
@@ -186,7 +186,10 @@ async def execute(self, session: Session, runner: Runner) -> Any:
             try:
                 # Deserialize parameters based on method signature
                 deserialized_params = _deserialize_params(method, self.params)
-                result = method(*deserialized_params)
+                async with session.keyword_lock:
+                    result = await asyncio.to_thread(method, *deserialized_params)
+                    heal_owner = getattr(method, "__self__", None)
```
thakur-patel, line 191: I'm curious as to why the healing part was moved to before the Exception is caught? Is it required as a part of this change or was it a general observation for that segment to be moved up the logic flow?

chinmayajha: before this PR as event loop was blocked, no race condition was possible (i.e. never 2 keywords at the same time). now that we offload to a thread we have to handle this special case, i.e. we have to manage `_last_heal_info` before the exception is caught i.e. the lock is released. 

We can move it out of `try:` as well. Basically a _last_heal_info management failure would not become a keyword execution failure in that case.

malto101 (approved):
LGTM

PR #461, opened 2026-08-20
refactor(action_keyword): drop located kwarg; explicit locate-and-act runner (#454)

### LINKED ISSUE #454: MCP/expose_api keyword dispatch: `located` treated as positional, breaks self-healing keywords when the element is found
## Summary

Element-interaction keywords that are decorated with `@with_self_healing`
(`press_element`, `enter_text`, and presumably every other locate-based
`ActionKeyword` method) fail with a `TypeError: ... takes from N to M
positional arguments but M+1 ... were given` when invoked through the
**named-params dispatch path** in `optics_framework/common/expose_api.py` —
i.e. via `optics mcp`, and likely via `optics serve`'s keyword-execution
endpoint too, since both appear to share this code path. Crucially, this
only manifests when a locate strategy actually **succeeds** in finding the
element; the same call against an element that isn't on screen fails with
a normal, correctly-reported "not found" error instead. That makes it easy
to miss in testing that only exercises failure paths, and it means the bug
surfaces exactly when the keyword would otherwise work.

Found while building an MCP demo against a real Android device (Appium +
UiAutomator2, real app, no mocks).

## Environment

- Commit: `e9613687fdb4267a293867e5889ffe157bda5f75` (origin/main, 2026-08-15)
- `optics --version`: 1.9.1
- Transport: `optics mcp --transport http`, called via `fastmcp.Client`
- Real device: Android 16 (OnePlus CPH2745) over USB, Appium 3.5.2 /
  UiAutomator2 driver 8.0.0

## Reproduction

```python
import asyncio, json
from fastmcp import Client

async def main():
    async with Client("http://127.0.0.1:8090/mcp") as client:
        session = await client.call_tool("start_session", {
            "driver": "appium",
            "url": "http://127.0.0.1:4723/wd/hub",
            "capabilities": {
                "platformName": "Android", "automationName": "UiAutomator2",
                "appPackage": "com.application.zomato", "appActivity": ".gold",
                "noReset": True,
            },
            "elements_sources": ["appium_find_element", "appium_page_source", "appium_screenshot"],
        })
        session_id = json.loads(session.content[0].text)["session_id"]

        await client.call_tool("launch_app", {"session_id": session_id})
        await asyncio.sleep(3)

        # Fails — element genuinely exists on screen at this point.
        await client.call_tool("enter_text", {
            "session_id": session_id,
            "element": '//*[contains(@text,"Search")]',
            "text": "shawarma",
        })

asyncio.run(main())
```

## Observed errors

`enter_text` (7 declared params: `element, text, aoi_x, aoi_y, aoi_width,
aoi_height, event_name`, plus `*, located`), called with only `session_id`,
`element`, `text` explicitly provided:

```
fastmcp.exceptions.ToolError: Execution failed: All fallback attempts failed for keyword 'enter_text':
('//*[contains(@text,"Search")]', 'shawarma', '0', '0', '100', '100') -> OpticsError('Code.E0701: Execution failed: Code.E0701: Execution failed: Code.E0401: Keyword execution failed: Code.X0201: All strategies failed for \'//*[contains(@text,"Search")]\' in \'enter_text\': ActionKeyword.enter_text() takes from 3 to 8 positional arguments but 9 positional arguments (and 1 keyword-only argument) were given')
```

`press_element` (9 declared params: `element, repeat, offset_x, offset_y,
index, aoi_x, aoi_y, aoi_width, aoi_height, event_name`, plus `*, located`),
same pattern:

```
fastmcp.exceptions.ToolError: Execution failed: All fallback attempts failed for keyword 'press_element':
('//*[contains(@text,"Search")]', '1', '0', '0', '0', '0', '0', '100', '100') -> OpticsError("Code.E0701: Execution failed: Code.E0701: Execution failed: Code.E0401: Keyword execution failed: Code.X0201: All strategies failed for '//*[contains(@text,\"Search\")]' in 'press_element': ActionKeyword.press_element() takes from 2 to 11 positional arguments but 12 positional arguments (and 1 keyword-only argument) were given")
```

In both cases the reported "takes from N to M" is exactly what the method's
own real signature allows — the overflow is always **exactly one argument
too many**.

By contrast, the identical `enter_text` call against a target that wasn't
on screen (app not yet launched) failed cleanly with `Code.E0201: No valid
strategies found for '...' in 'enter_text'` — no `TypeError`, no argument
overflow. So the bug is contingent on a locate strategy actually succeeding.

## Root cause (traced, not 100% certain but strongly supported by the code)

`optics_framework/common/expose_api.py:684`, in `_build_named_param_context`:

```python
sig = inspect.signature(method)
all_param_names = [p.name for p in sig.parameters.values() if p.name != "self"]
```

This only excludes `self` — it does **not** exclude `KEYWORD_ONLY`
parameters. Every `@with_self_healing`-decorated method ends its signature
with `*, located: Any = None` (e.g. `action_keyword.py:406`
`press_element(..., *, located: Any = None)`), so `located` ends up in
`all_param_names` alongside the real positional-or-keyword params.

Since `located` has a default (`None`) and is never present in the caller's
`params` dict, `_combo_to_positional_named` (`expose_api.py:718`) walks
`ctx.all_param_names` and appends `located`'s default value into the
positional-args list it builds for every combo — an extra slot that was
never meant to be positional at all:

```python
def _combo_to_positional_named(combo, ctx):
    combo_dict = dict(zip(ctx.provided_param_names, combo))
    out: List[str] = []
    for param_name in ctx.all_param_names:   # <-- includes "located"
        if param_name in combo_dict:
            out.append(combo_dict[param_name])
        elif param_name in ctx.param_defaults:  # <-- "located": None lands here
            ...
            out.append(default_val if isinstance(default_val, str) else str(default_val))
    return out
```

This `out` list is what ultimately gets passed down to the bound method as
positional arguments. Separately, `@with_self_healing`'s wrapper
(`action_keyword.py:142`) re-invokes the real function as:

```python
ret = func(self, element, located=result.value, *args, **kwargs)
```

`args` here is whatever the wrapper itself received beyond `element` — which
already contains the bogus extra "located" slot injected by
`_combo_to_positional_named` upstream. Combined with the wrapper's own
explicit `located=result.value` keyword, the real function ends up called
with one more positional value than its signature allows, hence the
`takes from N to M ... but M+1 were given` `TypeError`.

Notably, `optics_framework/helper/mcp_server.py`'s `_reflect_keyword_params`
already has the right idea for the **schema-generation** side of this — it
explicitly filters:

```python
_EXCLUDED_PARAMS = frozenset({"located"})
...
if name != "self" and name not in _EXCLUDED_PARAMS and p.kind not in (VAR_POSITIONAL, VAR_KEYWORD)
```

`expose_api.py`'s `_build_named_param_context` (and its sibling
`_resolve_named_to_positional`, which has the identical `if p.name != "self"`
filter at line ~598) never got the same treatment, so the exclusion exists
in the tool-schema path but not in the actual dispatch path.

## Suggested fix

In `_build_named_param_context` and `_resolve_named_to_positional`, exclude
`KEYWORD_ONLY` parameters (or specifically `located`, mirroring
`mcp_server._EXCLUDED_PARAMS`) from `all_param_names`/`param_names`, e.g.:

```python
all_param_names = [
    p.name for p in sig.parameters.values()
    if p.name != "self" and p.kind != inspect.Parameter.KEYWORD_ONLY
]
```

## Impact

Any MCP or (probably) `optics serve` caller driving an `ActionKeyword`
locate-based method by name — `press_element`, `enter_text`,
`press_checkbox`, `press_radio_button`, `swipe_from_element`,
`scroll_from_element`, `enter_number`, `clear_element_text`, `get_text`,
etc. — will hit this the moment the target element is actually found. This
makes the MCP entry point unreliable for the exact "AI agent taps/types on
a real screen" workflow it exists to support. `screenshot`, `launch_app`,
and session lifecycle calls are unaffected (confirmed working end-to-end
against a real device during this investigation).

chinmayajha:
## Summary

Fixes #454 — at the root, not the symptom.

Element-interaction keywords (`press_element`, `enter_text`, `get_text`, …) failed with `TypeError: … takes from N to M positional arguments but M+1 were given` through the **named-params dispatch path** (`optics mcp` tools and `optics serve`'s `/v1/sessions/{id}/action`) whenever a locate strategy **succeeded**. This PR removes the cause of that failure class entirely: the decorator-injected keyword-only `located` parameter is gone from every public keyword signature, replaced by an explicit locate-and-act runner.

## Root cause

Every locate-based keyword carried a `*, located: Any = None` tail param that only `@with_self_healing`'s wrapper ever supplied (it injected the located element/coordinates into the body). Because it sat in the public signature:

- the named-params dispatch in `expose_api.py` reflected it into a positional slot; `_combo_to_positional_named` stringified its `None` default into the args list, overflowing the bound call exactly when the element was found;
- `helper/mcp_server.py` needed a name-based `_EXCLUDED_PARAMS = {"located"}` filter to keep it out of tool schemas;
- `live.py` and `_heal_keyword_signature` each carried their own keyword-only skips;
- `/v1/keywords` still advertised it to HTTP clients;
- the wrapper's `located=` bypass existed solely so the deprecated `press_checkbox`/`press_radio_button` aliases could forward a located value without double-locating;
- AOI/index were read from kwargs only, so the all-positional convention MCP/serve uses silently ignored positional AOI values (confirmed empirically before this change).

Review feedback on the first iteration of this PR called the `located` design out: *a simple function can do it better*. Agreed — this PR implements that.

## What changed

**`optics_framework/api/action_keyword.py`**
- `@with_self_healing` deleted. Each locate-based keyword now builds an `act(located)` closure and calls the new `ActionKeyword._locate_and_act(element, func_name, act, intent_args, index, aoi)` runner.
- `_locate_and_act` + `_act_until_success` preserve the previous semantics exactly: screenshot capture → AOI parse/annotate → strategy ladder (XPath → text → OCR → image) → run `act` per result until one returns without raising → AI self-heal → `E0201`/`X0201` wrapping. Breadcrumbs (`_recent_steps`) unchanged.
- The 9 keywords (`press_element`, `press_checkbox`, `press_radio_button`, `swipe_from_element`, `scroll_from_element`, `enter_text`, `enter_number`, `clear_element_text`, `get_text`) have clean signatures. The two deprecated aliases plain-delegate to `press_element`.
- `index` is now an explicit parameter on the keywords that previously only accepted it via hidden wrapper kwargs; `get_text` gains explicit `index`/AOI params. **Behavior fix:** AOI/index are honored regardless of call convention — positional (MCP/serve) included.
- `_ai_self_heal` drops its unused `kwargs` parameter; stale comments updated.

**`optics_framework/helper/mcp_server.py`** — `_EXCLUDED_PARAMS` removed (dead once no signature carries `located`).

**`optics_framework/common/expose_api.py`** *(second commit, defense in depth)* — `_build_named_param_context` / `_resolve_named_to_positional` exclude `KEYWORD_ONLY` by kind, so the dispatch contract ("named params map to positional slots only") holds even if a future callable reintroduces a keyword-only param.

**Tests**
- New `TestKeywordSignatureHygiene`: no public API keyword on `ActionKeyword`/`AppManagement`/`Verifier`/`FlowControl` may declare `KEYWORD_ONLY`/`VAR_KEYWORD` params — the guard that would have caught #454's root cause at review time. (`FlowControl`'s deliberate `*args` keywords stay allowed.)
- New `TestLocateHonorsPositionalParams`: AOI/index reach the locator from both all-positional (MCP-style) and kwarg calls.
- Alias-delegation test rewritten for plain delegation; self-heal wiring tests updated to the new `_ai_self_heal` signature; MCP end-to-end tests bind all real self-healing signatures through both the dispatch path and actual MCP tool invocation.

## Verification

```
$ python -m pytest tests/ --no-cov -p no:warnings
736 passed, 1 skipped (pre-existing, playwright), 2 xfailed (pre-existing, #385)

$ ruff check optics_framework/ tests/
All checks passed!

$ pre-commit run --files <changed>
ruff/bandit/trailing-whitespace/end-of-file/check-json/gitleaks: Passed
```

## Impact

- Fixes #454 structurally: there is no keyword-only param left to mis-dispatch, for any entry point (MCP, serve, live, runner).
- MCP tool schemas and `/v1/keywords` now list only real, caller-suppliable parameters.
- AOI/index work over MCP/serve for the first time (previously silently ignored when sent positionally).
- No CSV/YAML/Robot-facing keyword names or existing positional call sites change; new params are appended with defaults, so existing callers are unaffected.

### DIFF
--- a/optics_framework/api/action_keyword.py
+++ b/optics_framework/api/action_keyword.py
@@ -1,4 +1,3 @@
-from functools import wraps
 import collections
 import inspect
 import shlex
@@ -39,16 +38,6 @@ def _parse_aoi_param(param: Any, default_value: float) -> float:
     if param is None or str(param).strip() in ('', 'None', 'none'):
         return default_value
     return float(param)
-
-
-def _parse_aoi_from_kwargs(kwargs: dict) -> Tuple[float, float, float, float, int, bool]:
-    aoi_x = _parse_aoi_param(kwargs.pop('aoi_x', '0'), 0)
-    aoi_y = _parse_aoi_param(kwargs.pop('aoi_y', '0'), 0)
-    aoi_width = _parse_aoi_param(kwargs.pop('aoi_width', '100'), 100)
-    aoi_height = _parse_aoi_param(kwargs.pop('aoi_height', '100'), 100)
-    index = int(kwargs.pop('index', 0))
-    is_aoi_used = not (aoi_x == 0 and aoi_y == 0 and aoi_width == 100 and aoi_height == 100)
-    return aoi_x, aoi_y, aoi_width, aoi_height, index, is_aoi_used
 
 
 def _maybe_save_aoi_screenshot(
@@ -110,64 +99,6 @@ def _save_annotated_for_result(
     )
 
 
-def _try_results_until_success(
-    results,
-    func: Callable,
-    self: Any,
-    element: Any,
-    args: tuple,
-    kwargs: dict,
-    screenshot_np: Any,
-    execution_dir: str,
-    func_name: str,
-):
-    last_exception = None
-    result_count = 0
-    locate_error: Optional[OpticsError] = None
-
-    try:
-        results_list = list(results)
-    except OpticsError as e:
-        # The locate generator raises when no strategy yields a result; treat that
-        # as "no results" so self-heal below gets a chance, but keep the original
-        # error as the cause instead of losing it.
-        internal_logger.debug(f"Locate generator raised for '{element}' in '{func_name}': {e}")
-        results_list = []
-        locate_error = e
-
-    for result in results_list:
-        result_count += 1
-        _save_annotated_for_result(result, screenshot_np, execution_dir, func_name)
-        try:
-            ret = func(self, element, located=result.value, *args, **kwargs)
-        except Exception as e:
-            internal_logger.error(
-                f"Action '{func_name}' failed with {result.strategy.__class__.__name__}: {e}")
-            last_exception = e
-            continue
-        # Success: record as a breadcrumb for any later AI self-heal (capture-then-return;
-        # appending after a bare `return` would be unreachable).
-        self._record_successful_step(func_name, element, args)
-        return ret
-    if result_count == 0:
-        if self._ai_self_heal(element, func_name, args, kwargs, screenshot_np):
-            return None
-        raise OpticsError(
-            Code.E0201,
-            message=f"No valid strategies found for '{element}' in '{func_name}'",
-            cause=locate_error,
-        ) from locate_error
-    if last_exception:
-        if self._ai_self_heal(element, func_name, args, kwargs, screenshot_np):
-            return None
-        raise OpticsError(
-            Code.X0201,
-            message=f"All strategies failed for '{element}' in '{func_name}': {last_exception}",
-            cause=last_exception,
-        )
-    raise OpticsError(Code.E0801, message=f"Unexpected failure: No results or exceptions for '{element}' in '{func_name}'")
-
-
 def _bounds_area(bounds: Optional[dict]) -> int:
     """
     indirect helper for `select_dropdown_option()`
@@ -218,33 +149,6 @@ def _raise_option_not_found(dropdown_element: str, option: str, interactive: Lis
             f"Available options: {available_texts}"
         ),
     )
-
-
-# Action Executor Decorator
-def with_self_healing(func: Callable) -> Callable:
-    @wraps(func)
-    def wrapper(self, element, *args, **kwargs):
-        # Skip self-healing if 'located' is already provided (avoids double healing)
-        if kwargs.get('located') is not None:
-            return func(self, element, *args, **kwargs)
-
-        screenshot_np = self._capture_screenshot_safe()
-        aoi_x, aoi_y, aoi_width, aoi_height, index, is_aoi_used = _parse_aoi_from_kwargs(kwargs)
-        if is_aoi_used and screenshot_np is not None:
-            _maybe_save_aoi_screenshot(
-                screenshot_np, aoi_x, aoi_y, aoi_width, aoi_height,
-                self.execution_dir, func.__name__,
-            )
-        self._save_screenshot_if_available(screenshot_np, f"pre-{func.__name__}")
-        results = _locate_element(
-            self.strategy_manager, element,
-            aoi_x, aoi_y, aoi_width, aoi_height, index, is_aoi_used,
-        )
-        return _try_results_until_success(
-            results, func, self, element, args, kwargs,
-            screenshot_np, self.execution_dir, func.__name__,
-        )
-    return wrapper
 
 
 class _HealProviders:
@@ -327,7 +231,7 @@ class ActionKeyword:
         self._recent_steps.append((func_name, [str(element), *map(str, args)]))
 
     def _ai_self_heal(
-        self, element: Any, func_name: str, args: tuple, kwargs: dict, screenshot_np: Any
+        self, element: Any, func_name: str, intent_args: tuple, screenshot_np: Any
     ) -> bool:
         """Last-resort AI recovery after every locate strategy failed. Returns True if healed.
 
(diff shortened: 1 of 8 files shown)


malto101:
i think it would be better to do away with located param and its related constructors in api_keywords
its unnecessarily convoluting it, a simple function can do it better

chinmayajha:
@malto101 This update should help

malto101:
Looks correct, could you run the unit tests once

thakur-patel (approved):

PR #462, opened 2026-08-20
fix(templatematch): bound per-call memory and fix index selection

chinmayajha:
## Summary

`TemplateMatchingHelper` allocated a fresh `cv2.SIFT_create` + `FlannBasedMatcher` on every `find_element`/`element_exist` call and ran uncapped SIFT over the full frame (~16-20MB/call in load testing). Replaced with a two-stage strategy; consolidating both public methods onto it also fixes three `index` defects.

### Memory
- **Fast path** — `cv2.matchTemplate` (`TM_CCOEFF_NORMED`): one bounded grayscale result map, no keypoints/descriptors/FLANN. Handles exact same-scale UI matches.
- **Fallback** — SIFT + FLANN + RANSAC, only when the fast path finds nothing. Reuses one feature-capped SIFT (`nfeatures=2000`, overridable via `capabilities.sift_nfeatures`) and one FLANN matcher per helper; releases descriptors before the homography; computes one transform (the old per-inlier loop produced N duplicate centers).

Singleton reuse is safe: the batch runner is sequential within a session and the serve/MCP path holds `Session.keyword_lock` (`execution.py:189`).

### `index` selection
1. **SIFT false positive on index miss.** `index=N` fell through to SIFT when the Nth distinct match was absent — even when real matches existed but fewer than `N+1`, and when a negative index wrapped via `peaks[-1]`. SIFT ignores `index` and returns its single best match → false-positive at the wrong spot. An `_INDEX_MISS` sentinel now separates "zero correlation matches" (SIFT-eligible) from "matches exist, this index selects none" (definitive miss); negative rejected.
2. **Per-axis suppression radius.** A single `max(tw, th)//2` radius merged stacked instances: a 14-row list with a full-row template (1000×90, r=500) collapsed to **3** peaks → `index=3..13` not-found for on-screen rows. Radii are now per-axis, floored at 1.
3. **Peaks in reading order.** `index=N` wasn't the Nth occurrence: three identical stamps at x=20/220/420 returned middle/third/first for 0/1/2. OCR engines index detection-order lists (easyocr.py:97, pytesseract.py:63, googlevision.py:66, remote_ocr.py:186), so `index` differed between text and image locators. Peaks now sort top-to-bottom, left-to-right; `index=None` still takes the strongest.

### Peak-extraction cost
Vectorised local-maximum prefilter (`res >= cv2.dilate(res, ...)`) so the greedy suppressor stops walking every plateau pixel. Matters because `index` defaults to `0` (not `None`) throughout the keyword layer (action_keyword.py:49 → strategies.py:348), so the loop runs on every image locate; `minMaxLoc` is only reachable from `element_exist`.

### Also
Restored `try/except cv2.error` around `cv2.perspectiveTransform`; dropped a dead confidence re-check after `minMaxLoc` and an unreachable `template is None` check (`load_template` raises internally); modernised typing to `X | None` / `tuple[...]` (Python 3.12+); trimmed narration to the non-obvious *why*.

## Test plan

| Case | Before | After |
|---|---|---|
| 14 list rows, full-row template | 3 peaks; `index=3..13` not-found | 14 peaks (y=200…1630); `index=14` → `_INDEX_MISS` |
| 3 stamps at x=20/220/420 | 0,1,2 → middle, third, first | 0,1,2 → x=20, 220, 420 |
| Single stamp | plateau → several | 1 peak; `index=1` → `_INDEX_MISS` |
| Synthetic 1080p gradient, 1.9M qualifying px | 16.4s | 0.85s |
| 2×-scaled stamp | — | fast path misses, SIFT resolves within 6px |

Automated: `tests/units/engines/vision_models/test_templatematch.py` (24 cases, synthetic ndarrays, no device):
- fast path: exact-match center/bbox, below-threshold miss, oversized/zero-sized template, `index=None` strongest
- `index` contract: reading-order selection, beyond-last → `_INDEX_MISS`, negative rejected
- `_distinct_peaks`: reading order regardless of score, plateau collapse, single-peak inverse, wide-short stacked-rows regression (8 survive), sub-radius merge limit
- `element_exist`: offset sign, missing-input `ValueError`, featureless-frame SIFT miss raises
- `find_element`: template files, `index=5`+2 matches → `None`, 2×-scaled SIFT within 6px
- `sift_nfeatures` default/override/coercion

- [x] `poetry run ruff check` — clean
- [x] `poetry run pre-commit run --files <changed>` — clean
- [x] `poetry run pytest tests/units` — 730 passed, 1 pre-existing skip, 2 pre-existing xfails

### Review follow-ups folded in
- Kept `cv2.error` detail + debug log on FLANN failure (was swallowed); `from e` chaining on all `RuntimeError` wraps
- Documented `confidence_level` dual semantics (corr threshold on fast path = stricter; Lowe ratio in SIFT = more lenient) and the sub-half-template merge limit in docstrings, both pinned with tests; `capabilities.sift_nfeatures` in `docs/configuration.md`
- perf: peak NMS bounded (boolean-map windows + `_PEAK_CANDIDATE_CAP` + stable sort) — ~2M qualifying px: 4.9s → 25ms, fuzz-verified identical survivors
- fix: `find_element`'s None-input error names the only triggering input; dropped impossible `Literal[False]` from `element_exist`'s return hint
- refactor: extracted `_apply_offset` (dedup'd `element_exist` offset tail), narrowed `_MatchResult`'s leading bool to `Literal[True]`, renamed stale `"SIFT assert_elements failed."`, annotated `_get_sift`/`_get_flann`
- test: `test_featureless_images_raise_from_sift_fallback` made robust to cv2 build differences (Linux wheels return 1.0 for identical solid regions, short-circuiting SIFT)

### Follow-up issues
- #472 — `confidence_level` inverted dual semantics; propose splitting into `corr_threshold` + `sift_ratio` with back-compat
- #471 — `confidence_level=1.0` can never match on the fast path; propose clamping below 1.0

### DIFF
--- a/optics_framework/engines/vision_models/image_models/templatematch.py
+++ b/optics_framework/engines/vision_models/image_models/templatematch.py
@@ -6,7 +6,7 @@ from optics_framework.common.logging_config import internal_logger
 from optics_framework.engines.vision_models.base_methods import load_template
 from optics_framework.common import utils
 
-_MatchResult = tuple[bool, tuple[int, int], tuple[tuple[int, int], tuple[int, int]]]
+_MatchResult = tuple[Literal[True], tuple[int, int], tuple[tuple[int, int], tuple[int, int]]]
 
 
 class _IndexMiss:
@@ -58,12 +58,12 @@ class TemplateMatchingHelper(ImageInterface):
         self._sift: cv2.SIFT | None = None
         self._flann: cv2.FlannBasedMatcher | None = None
 
-    def _get_sift(self):
+    def _get_sift(self) -> cv2.SIFT:
         if self._sift is None:
             self._sift = cv2.SIFT_create(nfeatures=self._sift_nfeatures)
         return self._sift
 
-    def _get_flann(self):
+    def _get_flann(self) -> cv2.FlannBasedMatcher:
         if self._flann is None:
             flann_index_kdtree = 1
             index_params = {"algorithm": flann_index_kdtree, "trees": 5}
@@ -300,8 +300,19 @@ class TemplateMatchingHelper(ImageInterface):
         if match_rule:
             return True, annotated_frame
 
-        internal_logger.warning("SIFT assert_elements failed.")
-        raise RuntimeError("SIFT assert_elements failed.")
+        internal_logger.warning("assert_elements failed.")
+        raise RuntimeError("assert_elements failed.")
+
+    @staticmethod
+    def _apply_offset(
+        center: tuple[int, int], top_left: tuple[int, int],
+        bottom_right: tuple[int, int], offset: list[int],
+    ) -> tuple[Literal[True], tuple[int, int], list[tuple[int, int]]]:
+        """Apply a pixel ``offset`` (x right, y up) and build the success tuple."""
+        center_x, center_y = center
+        center_x += offset[0]
+        center_y -= offset[1]
+        return True, (center_x, center_y), [top_left, bottom_right]
 
     def element_exist(
         self,
@@ -324,14 +335,10 @@ class TemplateMatchingHelper(ImageInterface):
 
         fast = self._match_template_fast(input_data, reference_data, confidence_level, None)
         if fast is not None:
-            _, (center_x, center_y), (top_left, bottom_right) = fast
-            center_x += offset[0]
-            center_y -= offset[1]
-            return True, (center_x, center_y), [top_left, bottom_right]
+            _, center, (top_left, bottom_right) = fast
+            return self._apply_offset(center, top_left, bottom_right, offset)
 
         center_x, center_y, top_left, bottom_right = self._sift_match(
             input_data, reference_data, confidence_level, min_inliers
         )
-        center_x += offset[0]
-        center_y -= offset[1]
-        return True, (center_x, center_y), [top_left, bottom_right]
+        return self._apply_offset((center_x, center_y), top_left, bottom_right, offset)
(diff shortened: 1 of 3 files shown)


chinmayajha:
@malto101 please review. This was done to reduce memory and CPU bursts.

optics_framework/engines/vision_models/image_models/templatematch.py
```diff
@@ -6,132 +6,285 @@
...
+            int(bbox_transformed[0][0][1]),
+        )
+        bottom_right = (
+            int(bbox_transformed[2][0][0]),
+            int(bbox_transformed[2][0][1]),
+        )
+
+        return center_x, center_y, top_left, bottom_right
+
+    def find_element(
```
malto101, line 238: When index=N but only N-1 matches exist (or fewer than N+1), the code loops through inliers computing centers/bboxes but stores only valid ones

malto101: assume If index=5 but only 2 distinct matches exist, the code raises IndexError instead of falling through. However, the caller then... has no choice but to retry differently. The real bug is that SIFT has no notion of index,it returns a single best match, so if the fast path fails and SIFT is consulted, it returns its best match at the wrong location, a silent false positive

chinmayajha: that's exactly the case _INDEX_MISS covers: index=5 with 2 peaks hits `index >= len(peaks)` and returns _INDEX_MISS before SIFT ever runs (templatematch.py:107-109), and `find_element` short-circuits to `None` on that instead of falling through to SIFT (templatematch.py:256). no IndexError, no silent SIFT false positive.

chinmayajha: @malto101

chinmayajha: this finding is there in the PR description

malto101 (approved):

PR #463, opened 2026-08-21
fix: reject fuzzy/partial xpath matches in presence assertions

### LINKED ISSUE #374: bug: `AppiumPageSource`'s `XPath` fuzzy fallback matching produces false positives for absent elements
When there are elements whose XPaths (resource-id/content-desc/etc.) are too similar to each other, querying for a non-existent element whose XPath is similar to those existing elements gets falsely reported as present.

This happens in `AppiumPageSource`'s XPath fallback matching (`find_xpath` in `optics_framework/engines/drivers/appium_UI_helper.py`), which uses fuzzy/partial matching as a last resort instead of requiring an exact match.

chinmayajha:
Fixes #374

## Summary

- `UIHelper.find_xpath` gains a `strict` flag: when set, resolution stops after the exact and relative (structure-only) stages, skipping the partial-`contains()` and fuzzy-attribute stages that loosen attribute-value comparison.
- `AppiumPageSource.assert_elements` (which backs `assert_presence` / visibility assertions) now resolves XPaths with `strict=True`, so an XPath whose attribute values are only partially or fuzzily matched on screen no longer reports the element as present.
- Interaction paths (`_locate_by_xpath`, `get_view_locator`) keep the default tolerant resolution, so click/tap self-healing behaviour is unchanged.

## Test plan

- [x] `poetry run pytest tests/units` — 718 passed
- [x] `poetry run ruff check` and pre-commit hooks pass on the touched files
- [x] Verified locally with `appium_page_source` enabled: asserting presence of `//android.widget.TextView[@content-desc="TRAI MySpeed"]` while only `content-desc="TRAI MySpeed Dashboard"` is on screen now times out instead of returning found; exact XPaths still assert successfully; `locate()` still resolves through partial/fuzzy matches so press/enter-text self-healing is unaffected.

### DIFF
--- a/optics_framework/engines/drivers/appium_UI_helper.py
+++ b/optics_framework/engines/drivers/appium_UI_helper.py
@@ -160,7 +160,7 @@ class UIHelper:
             return xpath
         return None
 
-    def find_xpath(self, xpath):
+    def find_xpath(self, xpath, strict: bool = False):
         """
         Process the given XPath and return the exact path from the UI tree after applying various matching strategies.
         """
@@ -193,27 +193,28 @@ class UIHelper:
                     f"Error in relative match for XPath '{xpath}': {str(e)}"
                 )
 
-            # 3. Partial Match
-            try:
-                found_xpath = self.find_partial(xpath)
-                if found_xpath:
-                    internal_logger.debug("Xpath found using partial match")
-                    return found_xpath, time_stamp
-            except Exception as e:
-                internal_logger.debug(
-                    f"Error in partial match for XPath '{xpath}': {str(e)}"
-                )
+            if not strict:
+                # 3. Partial Match
+                try:
+                    found_xpath = self.find_partial(xpath)
+                    if found_xpath:
+                        internal_logger.debug("Xpath found using partial match")
+                        return found_xpath, time_stamp
+                except Exception as e:
+                    internal_logger.debug(
+                        f"Error in partial match for XPath '{xpath}': {str(e)}"
+                    )
 
-            # 4. Attribute Match with Fuzzy Prefix and Suffix Handling
-            try:
-                found_xpath = self.find_attribute_match(xpath)
-                if found_xpath:
-                    internal_logger.debug("Xpath found using attribute match")
-                    return found_xpath, time_stamp
-            except Exception as e:
-                internal_logger.debug(
-                    f"Error in attribute match for XPath '{xpath}': {str(e)}"
-                )
+                # 4. Attribute Match with Fuzzy Prefix and Suffix Handling
+                try:
+                    found_xpath = self.find_attribute_match(xpath)
+                    if found_xpath:
+                        internal_logger.debug("Xpath found using attribute match")
+                        return found_xpath, time_stamp
+                except Exception as e:
+                    internal_logger.debug(
+                        f"Error in attribute match for XPath '{xpath}': {str(e)}"
+                    )
 
             # If no match is found
             internal_logger.debug(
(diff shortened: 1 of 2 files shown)


malto101:
Instead of pure rejection,
We should maybe have strict and relax flags that we can use

chinmayajha:
> Instead of pure rejection, We should maybe have strict and relax flags that we can use

You mean in the config? Wouldn't that be a bit unnecessary to overload the config with such a thing? I made this PR to highlight false positives on xpath matching. xpath shouldn't be fuzzy matched as something will almost always match. Can you suggest a better way to do this, if it exists? @malto101

malto101:
> > Instead of pure rejection, We should maybe have strict and relax flags that we can use
> 
> You mean in the config? Wouldn't that be a bit unnecessary to overload the config with such a thing? I made this PR to highlight false positives on xpath matching. xpath shouldn't be fuzzy matched as something will almost always match. Can you suggest a better way to do this, if it exists? @malto101

the main reason I dont want to remove it due to fact we are trying to make a no code framework. ideally, with limited info, it should be able proceed the location, 

maybe we could have another keyword like strict assert presence or something similar to that

chinmayajha:
> > > Instead of pure rejection, We should maybe have strict and relax flags that we can use
> > 
> > 
> > You mean in the config? Wouldn't that be a bit unnecessary to overload the config with such a thing? I made this PR to highlight false positives on xpath matching. xpath shouldn't be fuzzy matched as something will almost always match. Can you suggest a better way to do this, if it exists? @malto101
> 
> the main reason I dont want to remove it due to fact we are trying to make a no code framework. ideally, with limited info, it should be able proceed the location,
> 
> maybe we could have another keyword like strict assert presence or something similar to that

I definitely understand the intentrion, but in the past months of me using it, I have had numerous instances of false positives. xpath by design will have many characters matching, and will more often result in a false positive. Introducing a new keyword is fine, but I am questioning the accuracy of the old keyword itself, like why fuzzy matching on xpath if it consistently gives wrong matches

malto101:
> > > > Instead of pure rejection, We should maybe have strict and relax flags that we can use
> > > 
> > > 
> > > You mean in the config? Wouldn't that be a bit unnecessary to overload the config with such a thing? I made this PR to highlight false positives on xpath matching. xpath shouldn't be fuzzy matched as something will almost always match. Can you suggest a better way to do this, if it exists? @malto101
> > 
> > the main reason I dont want to remove it due to fact we are trying to make a no code framework. ideally, with limited info, it should be able proceed the location,
> > 
> > maybe we could have another keyword like strict assert presence or something similar to that
> 
> I definitely understand the intentrion, but in the past months of me using it, I have had numerous instances of false positives. xpath by design will have many characters matching, and will more often result in a false positive. Introducing a new keyword is fine, but I am questioning the accuracy of the old keyword itself, like why fuzzy matching on xpath if it consistently gives wrong matches

Got it!
But don't we have APPIUM find elements for that?

chinmayajha:
@malto101 yeah exactly, that's basically what strict mode gives you. Stage 1 of find_xpath runs your literal xpath against the ui tree, which is the same semantics as appium's own xpath lookup. Stage 2 keeps the structure but never loosens attribute values, that's what lets loose xpaths survive android/ios tag differences without matching wrong elements. The partial contains() and fuzzy ratio>=80 stages are our own self heal layer on top of the page source, they only exist so click/tap can still land somewhere close when the exact element isn't on screen. Assert presence has no business playing that game, hence this pr skips stages 3 and 4 for it.

On the separate keyword thought, I'd hold off for now since assert presence itself becomes the strict surface here, a second keyword doing the same thing would just confuse the no code story. And if we ever want strict interactions too, the flag already exists on find_xpath so wiring that up later is trivial.

malto101 (approved):

PR #488, opened 2026-08-26
fix(docs): API reference params + CLI layer architecture

chinmayajha:
### 1. API reference params not rendering
`/api_reference/action_keywords/...#ActionKeyword.swipe` showed the signature line but no parameter table. Root cause: the API classes use Sphinx-style docstrings (`:param x:`), but mkdocstrings defaulted to Google-style, so `:param:` lines rendered as raw description text.

- Set the mkdocstrings Python handler to `docstring_style: sphinx` so `:param:`/`:returns:`/`:raises:` parse into proper tables.
- Fixed the six docstrings the parser then surfaced:
  - `enter_text_using_keyboard`: `:param input:` -> `:param text_input:`
  - `launch_other_app`: `:param package_name:` -> `:param app_name:`
  - `assert_presence`: `:param timeout:` -> `:param timeout_str:`
  - `execute_script`, `get_interactive_elements`, `discover_templates`: dropped redundant `:type:` lines that duplicated the signature annotations.
- Converted the two Google-style docstrings in `flow_control.py` (`read_data`, `date_evaluate`) to Sphinx so they don't regress.

### 2. CLI Layer architecture not updated for the new commands
The `## Overview` list and the Mermaid diagram still named the six original commands and omitted `quickstart`, `doctor`, `configure`, `live`, and `mcp` (the Available Commands section below already had the onboarding ones). Updated both to the full command set, and added Live + MCP sections to Available Commands so the page lists every user-facing subcommand.

## Verification
- `mkdocs build` clean (no griffe warnings; only the pre-existing `mcp_usage.md` Docker-link warning remains, unrelated).
- `pytest tests/units/api/test_{flow_control_all,action_keyword,verifier,app_management}.py` -> 94 passed, 2 pre-existing xfailed.
- pre-commit green on all changed files.

## Commit layout
1. `fix(docs): render API reference params with sphinx docstring style` — mkdocs.yml + 5 API-class docstrings.
2. `docs: bring CLI layer architecture in line with the current commands` — `docs/architecture/cli_layer.md`.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -97,6 +97,8 @@ plugins:
       handlers:
         python:
           paths: [.]
+          options:
+            docstring_style: sphinx
   - minify:
       minify_html: true
 
--- a/docs/architecture/cli_layer.md
+++ b/docs/architecture/cli_layer.md
@@ -6,12 +6,14 @@ The Optics Framework provides a command-line interface (CLI) that enables users 
 
 The CLI layer provides:
 
-1. **Project Management** - Initialize and manage test projects
-2. **Test Execution** - Execute and validate test cases
-3. **Driver Setup** - Install and configure drivers
-4. **Code Generation** - Generate test framework code
-5. **API Server** - Start the REST API server
-6. **Configuration** - Manage framework configuration
+1. **Onboarding** - `optics quickstart`, `optics doctor`, and `optics configure` take a new user from a fresh install to a runnable project and a healthy setup
+2. **Project Management** - `optics init` scaffolds a project from a template
+3. **Test Execution** - `optics execute` and `optics dry_run` run and validate test cases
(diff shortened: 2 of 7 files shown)


mkdocs.yml
```diff
@@ -97,6 +97,8 @@ plugins:
       handlers:
         python:
           paths: [.]
+          options:
+            docstring_style: sphinx
```
thakur-patel, line 101: Why did this have to be changed? Also, could you confirm that other components aren't affected by this?

chinmayajha: mkdocstrings' Python handler defaults to Google-style docstring parsing
the 5 API classes are in sphinx style parsing which didn't render because of that

And yes I have rendered it locally, other components won't be affected

thakur-patel (approved):

PR #489, opened 2026-08-31
fix: reject fuzzy/partial text matches in presence assertions

chinmayajha:
## Summary

- `AppiumPageSource.assert_elements` already resolved **XPath** locators strictly (d844c98); **text** locators had no equivalent guard. A presence/"element exists" check could report `true` for text that only vaguely resembled what was actually on screen -- `compare_text`'s `fuzz.ratio` fallback scores `"Buy 102 devices"` at 93 against an on-screen `"Buy 101 devices"`. In App Studio, a Condition block probing for the former silently succeeded, so its ELSE branch could never run.
- `compare_text` gains a `strict` param that skips the partial/fuzzy stages (exact match only).
- `UIHelper.get_locator_and_strategy` / `get_locator_and_strategy_using_index` gain the same `strict` param, mirroring `find_xpath`'s existing strict mode for XPath locators.
- `AppiumPageSource.assert_elements` now always resolves text strictly -- this alone fixes the reported bug, unconditionally.
- New `Config.strict_element_match` (default `False`) threads the same flag through `AppiumPageSource`'s *locate* paths (`_locate_by_xpath`, `_locate_by_text`), so a project can opt in to exact-only matching for tap/self-heal locates too. Locate stays fuzzy-tolerant by default, per d844c98's original intent (self-healing resilience).

## Test plan

- [x] `poetry run pytest tests/` -- full suite green (2 pre-existing unrelated xfails)
- [x] `poetry run ruff check` / `poetry run pre-commit run --files <changed>` -- clean
- [x] New unit tests cover: `compare_text` strict vs fuzzy (incl. the exact "Buy 102/101 devices" regression), `UIHelper` strict-mode locator resolution, `AppiumPageSource` presence-assertion strictness and locate-path config wiring, `Config.strict_element_match` default


thakur-patel (approved):
Double check to update the docs

PR #501, opened 2026-08-31
fix: break infinite retry loop on colliding quickstart project name

chinmayajha:
## The bug

`optics quickstart` hangs forever when the default project name already exists on disk and the user answers the rename prompt by pressing Enter.

Reproduced end-to-end against the real `rich.Prompt.ask` (collaborators stubbed, real stdin fed by `yes ""`, all other prompts left alone): the wizard emitted **201 prompts** before I cut it off, alternating

```
Project '/tmp/.../my-optics-project' already exists. Choose a different name.
Project name: A project name can't be empty.
Project '/tmp/.../my-optics-project' already exists. Choose a different name.
Project name: A project name can't be empty.
...
```

Only Ctrl-C gets out. There is no auto-suggested alternative and empty input is not an escape hatch.

## Root cause

Two things combine in `run_quickstart`:

1. The collision re-prompt was `Prompt.ask("Project name")` — **no `default=`**. Rich's `PromptBase.__call__` only substitutes a default when one was passed (`if value == "" and default != ...: return default`); otherwise `process_response("")` returns `""` cleanly, no `InvalidResponse`, no internal re-prompt. So the empty answer comes straight back into our code. The loop is entirely ours, not Rich's.
2. The empty answer hit `if not name: ...; continue`, which jumped back to `while os.path.exists(project_path):` **without recomputing `project_path`**. The condition was therefore trivially still true, forever.

## The fix

The moment a collision is detected, compute the first free `name-N` under the chosen base path and offer it as the re-prompt's `default`. Pressing Enter now always advances to an available path, so the loop terminates. Same repro after the fix: 5 prompts total, project created at `my-optics-project-2`.

That makes the "can't be empty" branch unreachable, so it is deleted rather than left as dead code. A whitespace-only answer (which Rich strips to `""` even when a default exists) is folded onto the same forward-progress path by `or suggestion`. The `EOFError` → exit 1 handling for genuinely non-interactive/piped-without-terminal contexts is unchanged.

## Blast radius

Checked the other "retry until a free name/path" sites — the defect is unique to `quickstart.py`:

- `initialize._check_and_prepare_directory` raises `ValueError` on collision (one-shot, `cli.py` maps it to exit code 3) — not a loop.
- `initialize._pick_template`'s `while True` treats empty input as "cancel" and returns `None` — has an escape hatch.
- `config_manager.py:307` is a `Confirm.ask`, not a loop.

## Testing

- Two regression tests in `TestExistingProject`, driven by an `AlwaysEnter` stub that faithfully emulates Rich (returns the question's `default` on empty input, `""` when there is none) with a 20-prompt cap so a regression fails fast instead of hanging CI. One asserts the wizard lands on `my-optics-project-2`; the other that the suggestion skips suffixes already taken (`-2`/`-3` occupied → `-4`). **Both fail on the pre-fix code with `AssertionError: prompt loop did not terminate within 20 questions`** and pass after.
- Full `tests/units/helpers/` suite green (454 tests).
- `pre-commit run --files ...` — ruff, bandit, whitespace, EOF, gitleaks all pass.


thakur-patel (approved):
LGTM

PR #503, opened 2026-08-31
fix: catch ImportError at CLI startup instead of a raw cv2/libGL traceback

chinmayajha:
## The bug

On a minimal or headless Linux box (fresh container, CI runner) with no `libGL.so.1`, the bare `optics` command and `optics --help` die with a raw traceback ending in:

```
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```

`optics_framework/common/utils.py:6` imports `cv2` at module level, and OpenCV needs libGL. The fix is named nowhere — not in the crash output, not in the README, not in the docs (`grep -rn "libGL\|libgl" docs/ README.md` returns nothing). A fresh install on such a box is a dead end.

## Root cause — not where it looked

The crash does **not** originate in `cli.py`'s own import block. Blocking `cv2` with a meta-path finder and importing each of `cli.py`'s 13 helper imports individually shows all of them failing with the *same* 23-frame chain, always starting above `cli.py`:

```
optics_framework/__init__.py:10   from optics_framework.optics import Optics
optics_framework/optics.py:29     from optics_framework.api.app_management import AppManagement
optics_framework/api/__init__.py:1
optics_framework/api/action_keyword.py:9  from optics_framework.common.strategies import StrategyManager
optics_framework/common/strategies.py:9   from optics_framework.common import utils
optics_framework/common/utils.py:6        import cv2
```

The package `__init__.py` eagerly re-exported `Optics`, so *every* import of *any* `optics_framework` submodule dragged in the whole API and vision stack. That runs while Python resolves the console-script target `optics_framework.helper.cli` — **before `cli.py` executes a single line**. A `try/except` inside `cli.py` alone would have been dead code for this bug.

Same root cause hits `from optics_framework import Optics` (the SDK/Robot facade), at 15 frames.

## The fix

Two halves, both required:

1. **`optics_framework/__init__.py`** — resolve the `Optics` re-export lazily (PEP 562 `__getattr__`). Importing a submodule no longer pulls the API/vision stack, which is what makes a guard in `cli.py` reachable at all. `from optics_framework import Optics` still works; `__dir__` keeps it discoverable; a `TYPE_CHECKING` import keeps type checkers happy.
2. **`optics_framework/helper/cli.py`** — wrap the helper import block in `try/except ImportError` and render the failure through the same rich `Console`/`Panel` style `execute.py:_abort_preflight` already uses, then `sys.exit(1)`.

The message **always prints the real exception text**, so nothing is swallowed or reinterpreted. It adds the concrete package name only for the libGL case (`apt-get install -y libgl1` / `dnf install -y mesa-libGL` / `apk add mesa-gl`), and a reinstall hint otherwise.

```
╭──────────────────────────────── Cannot start ────────────────────────────────╮
│ Optics could not load one of its dependencies, so no command can run.        │
│                                                                              │
│ Underlying error: libGL.so.1: cannot open shared object file: No such file   │
│ or directory                                                                 │
│                                                                              │
│ OpenCV needs a system graphics library that minimal and headless             │
│ Linux images do not ship. Install it, then re-run the command:               │
│                                                                              │
│   Debian/Ubuntu:  sudo apt-get install -y libgl1                             │
│   RHEL/Fedora:    sudo dnf install -y mesa-libGL                             │
│   Alpine:         apk add mesa-gl                                            │
╰──────────────────────────────────────────────────────────────────────────────╯
```

### Scope

This makes the failure legible; it does **not** make `optics --help` run without OpenCV. 13 modules import `cv2` at top level — removing that dependency is a much larger, riskier change than this bug warrants, and is deliberately out of scope here.

## Testing

- **Automated regression tests** in `tests/units/helpers/test_cli.py` (`TestStartupImportGuard`). Two pin the message contract (libGL names the installable package; an unrelated `ImportError` still shows its real text and does *not* get a graphics hint). A third reproduces the crash end to end in a subprocess with a meta-path finder that fails `cv2` exactly the way a missing `libGL.so.1` does — a fresh interpreter is required both because the parent already has `cli` imported and because only that exercises the real console-script import order. It asserts the package still imports, then that importing `cli` exits 1 with the hint and no `Traceback`.
- **Before/after verified by simulation.** Same `cv2` block against `main` reproduces the raw traceback; against this branch it prints the panel and exits 1.
- **Happy path, unchanged:** `optics --help` renders the full subcommand list and exits 0; bare `optics` still shows the onboarding panel (run against an isolated `OPTICS_HOME`).
- **Facade unchanged:** `from optics_framework import Optics` resolves, `'Optics' in dir(optics_framework)` is true, and an unknown attribute still raises `AttributeError`. `tests/units/test_optics.py` explicitly guards this documented import and passes.
- **Full suite:** `pytest` — only `TestExtractionCost::test_cost_growth_is_linear_not_quadratic` fails, and it passes in isolation. It is a wall-clock ratio test whose own docstring cites "timing noise on shared CI"; it is load-sensitive and unrelated to import wiring.
- `pre-commit run --files` on all three changed files: ruff, bandit, whitespace, EOF, gitleaks all pass.

## Review follow-ups

**Wildcard export (P1).** The first cut left `__all__` empty, which silently dropped `Optics` from `from optics_framework import *`. Restored. `__all__` only matters to a star import, and a star import asks for the facade by name — resolving it eagerly there is the documented contract, identical to `from optics_framework import Optics`. Plain `import optics_framework` and every submodule import stay lazy, which is the half that keeps the console script reaching the guard. The cv2-blocking subprocess test asserted the opposite and now pins the corrected behaviour.

The export surface had no coverage in a lean environment — `tests/units/test_optics.py` is `importorskip`'d on easyocr, which is how this got through. The contract moved to `tests/units/test_package_exports.py`, which always runs, and now covers the wildcard case too.

**CLAUDE.md anchors (P2).** Refreshed for `cli.py` and `execute.py` — the two files this PR shifts. All 23 rewritten anchors were verified to land on the symbol the prose names. Most of the correction is pre-existing drift rather than this change (`main()` was cited at `:390` and sits at `:529`; `BaseRunner` at `:492`, now `:679`). Also corrects `find_files(folder_path, validate=True)` — there is no `validate` argument — and documents the import guard in step 1.

Anchors for files this PR does not touch (`test_runnner.py`, `strategies.py`, `live.py`, `optics.py`, …) are still stale and are left for a separate docs pass.

**Rebased** onto `main`; it was 78 commits behind, so anchors computed against the old base would have been wrong on merge.

<h2><a href="https://app.greptile.com/api/retrigger?id=65923299"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because it breaks the existing wildcard export of `Optics`, and the repository-required CLI anchor updates must also be made.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Aoptics_framework%2F__init__.py%3A23%0ABefore%20this%20change%2C%20%60from%20optics_framework%20import%20*%60%20bound%20the%20documented%20%60Optics%60%20facade.%20Because%20%60__all__%60%20is%20now%20empty%2C%20the%20same%20import%20exports%20nothing%2C%20so%20downstream%20references%20to%20%60Optics%60%20fail%20with%20%60NameError%60.%20Keeping%20%60Optics%60%20in%20%60__all__%60%20preserves%20the%20existing%20wildcard-import%20contract%20while%20ordinary%20package%20and%20submodule%20imports%20remain%20lazy.%0A%0A%60%60%60suggestion%0A__all__%3A%20list%5Bstr%5D%20%3D%20%5B%22Optics%22%5D%0A%60%60%60%0A%0A%23%23%23%20Issue%202%0Aoptics_framework%2Fhelper%2Fcli.py%3A9-37%0AThis%20startup%20block%20shifts%20every%20following%20CLI%20definition%20without%20updating%20the%20corresponding%20%60CLAUDE.md%60%20runtime-map%20anchors%2C%20including%20the%20entry%20for%20%60main%28%29%60.%20The%20repository%20directive%20requires%20mismatched%20%60path%3Aline%60%20anchors%20to%20be%20fixed%20rather%20than%20trusted.%20These%20references%20must%20be%20updated%20before%20merging%20so%20maintainers%20are%20not%20sent%20to%20the%20wrong%20code%20locations.%0A%0ANote%3A%20If%20this%20suggestion%20doesn't%20match%20your%20team's%20coding%20style%2C%20reply%20to%20this%20and%20let%20me%20know.%20I'll%20remember%20it%20for%20next%20time!%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=mozarkai%2Foptics-framework&pr=503&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Wildcard Export Drops Optics** <a href="https://github.com/mozarkai/optics-framework/pull/503#discussion_r4047203059">▶</a>
2. P2&nbsp;**CLI Anchors Remain Stale** <a href="https://github.com/mozarkai/optics-framework/pull/503#discussion_r4047203083">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
optics_framework/__init__.py:23
Before this change, `from optics_framework import *` bound the documented `Optics` facade. Because `__all__` is now empty, the same import exports nothing, so downstream references to `Optics` fail with `NameError`. Keeping `Optics` in `__all__` preserves the existing wildcard-import contract while ordinary package and submodule imports remain lazy.

```suggestion
__all__: list[str] = ["Optics"]
```

### Issue 2
optics_framework/helper/cli.py:9-37
This startup block shifts every following CLI definition without updating the corresponding `CLAUDE.md` runtime-map anchors, including the entry for `main()`. The repository directive requires mismatched `path:line` anchors to be fixed rather than trusted. These references must be updated before merging so maintainers are not sent to the wrong code locations.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<details><summary>Summary</summary>

This PR makes package initialization lazy so the CLI can intercept dependency import failures, adds shared startup-abort rendering with libGL-specific guidance, and adds regression coverage.

- The missing-libGL startup failure is real and the guard is needed on the declared console-script path.
- The shared abort helper preserves existing execute preflight behavior while adding a plain-text fallback.
- The implementation introduces a package export compatibility regression by removing `Optics` from `__all__`.
- Changed CLI line locations also require the repository's runtime-map anchors to be refreshed.
</details>

<details><summary>Diagram</summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[optics console script] --> B[Import optics_framework package]
    B --> C[Lazy package initialization]
    C --> D[Import helper.cli]
    D --> E[Import CLI helper modules]
    E --> F{Dependency ImportError?}
    F -- No --> G[Parse and dispatch command]
    F -- libGL --> H[Render libGL installation guidance]
    F -- Other --> I[Render underlying error and reinstall guidance]
    H --> J[Exit 1 without traceback]
    I --> J
    K[from optics_framework import Optics] --> L[__getattr__ resolves facade]
    M[from optics_framework import *] --> N[Empty __all__ omits Optics]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["test: cover the CLI startup import guard"](https://github.com/mozarkai/optics-framework/commit/1f442ae876ae5fe73ded04ca82bfd51a42d207ce)</sub>

### DIFF
--- a/optics_framework/__init__.py
+++ b/optics_framework/__init__.py
@@ -5,8 +5,31 @@ It is re-exported here so the documented import works::
 
     from optics_framework import Optics
 
+The re-export is resolved lazily (:pep:`562`). Importing it eagerly would pull
+the whole API and vision stack — including ``cv2`` — into *every* import of any
+``optics_framework`` submodule, so a missing ``libGL.so.1`` crashed the
+``optics`` console script before ``helper/cli.py`` could run a single line.
+
+There is deliberately no star-export surface: a star import resolves every
+name in ``__all__`` through :func:`__getattr__`, which would eager-import
+``Optics`` and resurrect that crash on machines without the graphics library.
 """
 
-from optics_framework.optics import Optics
+from typing import TYPE_CHECKING, Any
 
-__all__ = ["Optics"]
+if TYPE_CHECKING:
+    from optics_framework.optics import Optics  # noqa: F401 - type-checker-only re-export for the lazy facade
+
+__all__: list[str] = []
+
+
+def __getattr__(name: str) -> Any:
+    if name == "Optics":
+        from optics_framework.optics import Optics
+
+        return Optics
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    return sorted([*globals(), "Optics"])
--- a/optics_framework/helper/abort.py
+++ b/optics_framework/helper/abort.py
@@ -1,0 +1,60 @@
+"""Friendly startup-abort rendering shared by cli.py and execute.py."""
+
+import sys
+from dataclasses import dataclass
+from typing import NoReturn
+
+_LIBGL_MARKERS = ("libGL", "libgl")
+
+_GRAPHICS_LIBRARY_GUIDANCE = (
+    "OpenCV needs a system graphics library that minimal and headless",
+    "Linux images do not ship. Install it, then re-run the command:",
+    "",
+    "  Debian/Ubuntu:  apt-get install -y libgl1",
+    "  RHEL/Fedora:    dnf install -y mesa-libGL",
+    "  Alpine:         apk add mesa-gl",
+)
+
+_REINSTALL_GUIDANCE = (
+    "Reinstall the package and its dependencies, then re-run:",
+    "",
+    "  pip install --force-reinstall optics-framework",
+)
+
+
+@dataclass(frozen=True)
+class ImportFailure:
+    """A startup import error paired with the guidance that explains it."""
+
+    error: ImportError
+    guidance: tuple[str, ...]
+
+
+def classify_import_error(exc: ImportError) -> ImportFailure:
+    """Pair an import error with its fix guidance.
+
+    OpenCV reports a missing system graphics library as a message substring
+    rather than a structured attribute, so the markers are the only signal
+    that survives across OpenCV versions.
+    """
+    if any(marker in str(exc) for marker in _LIBGL_MARKERS):
+        return ImportFailure(exc, _GRAPHICS_LIBRARY_GUIDANCE)
+    return ImportFailure(exc, _REINSTALL_GUIDANCE)
+
+
+def abort_with_panel(lines: list[str]) -> NoReturn:
+    """Print ``lines`` in a red "Cannot start" panel on stderr, then exit 1.
+
+    rich is imported lazily: when the install is broken enough that rich
+    itself is missing, the caller must still get the guidance, not a
+    second traceback.
+    """
+    try:
+        from rich.console import Console
+        from rich.panel import Panel
+
+        Console(file=sys.stderr).print(
+            Panel("\n".join(lines), title="Cannot start", border_style="red"))
+    except ImportError:
+        print("\n".join(lines), file=sys.stderr)
+    sys.exit(1)
--- a/optics_framework/helper/cli.py
+++ b/optics_framework/helper/cli.py
@@ -1,21 +1,40 @@
 import argparse
 import os
 import sys
-from typing import Literal, Optional
-from pydantic import BaseModel
-from optics_framework.helper.list_keyword import main as list_main
-from optics_framework.helper.initialize import create_project, available_templates
-from optics_framework.helper.version import VERSION
-from optics_framework.helper.execute import execute_main, dryrun_main
-from optics_framework.helper.live import live_main
-from optics_framework.helper.generate import generate_test_file as generate_framework_code
-from optics_framework.helper.setup import EngineInstallerApp, SetupError, list_engines, install_extras, resolve_engines, print_install_next_steps
-from optics_framework.helper.serve import run_uvicorn_server
-from optics_framework.helper.autocompletion import update_shell_rc
-from optics_framework.helper.config_manager import configure as configure_project
-from optics_framework.helper.onboarding import welcome, is_first_run, mark_onboarded
-from optics_framework.helper.doctor import run_doctor
-from optics_framework.helper.quickstart import run_quickstart
+from typing import Literal, NoReturn, Optional
+
+from optics_framework.helper.abort import abort_with_panel, classify_import_error
+
+
+def _abort_on_import_error(exc: ImportError) -> NoReturn:
+    """Report a startup import failure as guidance instead of a traceback."""
+    failure = classify_import_error(exc)
+    abort_with_panel([
+        "Optics could not load one of its dependencies, so no command can run.",
+        "",
+        f"Underlying error: {failure.error}",
+        "",
+        *failure.guidance,
+    ])
+
+
+try:
+    from pydantic import BaseModel
+    from optics_framework.helper.list_keyword import main as list_main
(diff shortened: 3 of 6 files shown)


thakur-patel (approved):
Approving as this is the scope of effect is minimal to catching an error during optics-framework setup

PR #505, opened 2026-08-31
fix: stop optics live's /help panel from silently truncating

chinmayajha:
## The bug

In `optics live`, `/help` opened a popup that silently cut off part-way through
the slash-command table. `/device`, `/quit` and the entire **Keys** section were
never visible — at any terminal size, with no scrollbar, no "more below" hint and
no key that could reach them.

Verified in `optics_framework/helper/live_tui.py`:

- The popup `Float` hardcoded `height=Dimension(min=1, max=30)`, and unlike the
  keyword browser (`overlay_float`, which has `up`/`down` bindings gated on
  `overlay_active` and a cursor the window scrolls to) the popup had no scroll
  bindings at all — the window always rendered from row 0 and clipped the rest.
- `_HELP_TEXT` is 41 authored lines, which is **51 rendered rows** once wrapped to
  the popup's 72-column preferred width. Row 31 lands on
  `  /device [id]   List/switch connected Android + iOS devices …`, so everything
  from `/device` down was unreachable — matching the report exactly.
- `/elements` shares the same popup, so any project with more than ~14 named
  elements (2 rows each) lost its tail the same way. That is the only other
  instance: `overlay_float`'s `max=18` is fine because its control reports a cursor
  position, which prompt_toolkit scrolls to keep visible.

## The fix

Took resolution **(a), real scrolling** — the file already had the binding pattern
to copy, and dropping the cap alone would still clip on a terminal shorter than the
content:

- `_render_popup()` renders a window over the popup text starting at
  `popup_scroll`, with a footer showing the visible range
  (`↑/↓ PgUp/PgDn scroll · 1-29/51`) whenever the content overflows. Short popups
  render unchanged, with no footer.
- `up`/`down` (a row) and `pageup`/`pagedown` (a page) are bound behind
  `Condition(lambda: self.popup_active)`, mirroring the overlay's `overlay_on`
  bindings. The existing history-recall bindings were already excluded while a
  popup is open, so nothing else changed.
- The viewport is measured from the live terminal size (capped at 30 rows /
  72 columns) instead of assuming 30 rows exist, so a short terminal just shows a
  smaller page rather than hiding its tail; and the text is pre-wrapped to the
  popup width so the row arithmetic matches what prompt_toolkit draws.
- Docs (`docs/usage/live_usage.md`) and the in-app Keys list mention the scroll keys.

## Testing

`tests/units/helpers/test_live_keywords.py` already drives a real `LiveTUI` against
a stubbed prompt_toolkit app; six tests were added there in the same style:

- previously-clipped markers (`/device`, `/quit`, `Keys`, `Ctrl-C`,
  `Press Esc to close this help.`) are absent from page 1 and present after one
  page-scroll;
- the footer reports the visible range;
- scrolling clamps at both ends and the last line is reachable at the bottom;
- on a 12-row terminal the page shrinks to 9 rows and the tail is still reachable;
- the `up`/`down`/`pageup`/`pagedown` bindings resolve (exactly one each) while the
  popup is open and move the offset — this is the regression test for the wiring
  that was missing;
- a short popup renders with no indicator and byte-identical text.

`pytest tests/units` passes (only the pre-existing `#385` XFAILs), and
`pre-commit run --files …` is clean on all three changed files.

### DIFF
--- a/docs/usage/live_usage.md
+++ b/docs/usage/live_usage.md
@@ -86,9 +86,9 @@ sleep 5
 |------------------|-------------|
 | `/save <test_case> <module_name>` | Save the recorded actions as `<module_name>` in `modules/modules.csv` and add a `(<test_case>, <module_name>)` row to `test_cases/test_cases.csv`. Elements are merged into your project's existing elements CSV when one is tracked anywhere (the templates keep it at `test_data/elements.csv`) — only names not already present are appended, so no second `elements/elements.csv` appears; with no elements file, a header-only `elements/elements.csv` stub is created. All three files use fixed names and are **appended** to when they already exist, so you can build a suite one module at a time. A successful save **clears the recording buffer**, so the next actions become the next module. If `<module_name>` or `<test_case>` already exists, the save is refused with a prompt — re-run the identical `/save` to append, or pick a different name. Session screenshots/artifacts are also snapshotted to `execution_output/<module_name>/`. |
 | `/device [id]`   | **Appium sessions only.** List the **Android** (`adb`) and **iOS** (`idevice_id`) devices attached to *this* machine, each labelled by platform; with no argument, pick one to switch the active device's `udid`. The chosen device must match the session's configured platform. For Selenium/Playwright it reports that switching doesn't apply (the target is the configured browser). See the note below for remote Appium hubs. |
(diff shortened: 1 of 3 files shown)


thakur-patel (approved):
LGTM, no core modules are affected here

PR #506, opened 2026-08-31
fix: surface a clear error for a dead driver session instead of masking it

chinmayajha:
## The bug

A real-device session against a remote Appium hub went idle for 5-10 minutes (well under `newCommandTimeout`) and the hub dropped it server-side. Every keyword afterwards failed with an error that never mentions a session:

```
[E0201] element not found            <- locate / assert_presence
[E0202] No interactive elements retrieved using available strategies.
[E0403] No pagesource captured using available strategies.
InvalidSessionIdException: the session is not running   <- the actual cause, unwrapped
```

Separately, re-running `launch_app` -- the keyword that created the session, ~12s originally -- returned `launch_app 0.00s` as a success. Nothing had been launched; the next `/screenshot` failed identically.

## Root cause

**1. The strategy layer reinterprets any backend exception as "not found".** Every entry point in `common/strategies.py` wraps its backend call in a broad `except Exception` and converts the failure into an element/pagesource/screenshot miss. `InvalidSessionIdException` hits those handlers and comes out as `E0201`/`E0202`/`E0303`/`E0403`. There was **no handling of `InvalidSessionIdException` anywhere in `optics_framework/`** before this PR (repo-wide grep: zero matches).

The innermost one matters most: `LocatorStrategy._assert_elements_locator_style` *returns* `(False, None, None)` rather than raising, so on the assert path the exception never even reached the handlers above it.

**2. `Appium.launch_app` trusted a client-side cached id.** It gated on `self.driver is None`; if the client object existed it returned `driver.session_id` and did nothing. `driver.session_id` is a local attribute that outlives the server-side session, so against a dead session this is a false-positive success in ~0s.

## The fix

| File | Change |
|---|---|
| `common/error.py` | New `Code.E0106` + `ErrorSpec` ("Driver session is no longer active", category `DRIVER`, 500) |
| `common/session_liveness.py` *(new)* | `is_dead_session_error` / `raise_if_session_dead`; guarded selenium import |
| `common/strategies.py` | `raise_if_session_dead(e)` in the 8 confirmed masking handlers |
| `engines/drivers/appium.py` | `_is_session_alive()` probe; `launch_app` restarts a dead session |

**Why `E0106` and not `E02xx`/`X0201`:** those prefixes drive the param and strategy fallback ladders (`test_runnner.py:497`/`:1121`, `helper/live.py:654`). Retrying other locators or element candidates can never revive a dead session, so it must fail fast. `E0106` follows `E0105`, the existing "fatal, not part of the `E02*` family" driver-layer code.

`is_dead_session_error` walks both `__cause__` and `__context__`, because the layers below re-raise in both styles -- `appium_find_element.py` wraps with `cause=e`, `ScreenshotStrategy.capture` re-raises bare inside an `except` block -- so the original exception is reachable only via one link or the other.

The eight `strategies.py` sites are the ones that demonstrably mask this today, not defensive additions:

```
_assert_elements_locator_style       swallowed into (False, None, None)
_try_strategy_locate                 -> locate() E0201
_assert / _try_assert_with_strategy  -> E0201
capture_screenshot                   -> E0303
capture_screenshot_bytes             -> E0303
capture_pagesource                   -> E0403
get_interactive_elements             -> E0202
```

`_is_session_alive` treats only `invalid session id` as conclusive. Any other failure -- an unsupported window-rect command on a TV profile, a transient network error -- leaves the session presumed alive, so a working session is never torn down underneath the caller.

## Blast radius checked

- **Selenium driver** shares the masking (same `InvalidSessionIdException`, `WebDriverException` subclass) and is covered for free, since the guards sit in the driver-agnostic strategy layer. But its `launch_app` does **not** share the false positive: it always issues a real `driver.get()`, which raises on a dead session and is re-raised. Left alone.
- **Playwright** uses a different hierarchy (`playwright.sync_api.TimeoutError` etc.), no `InvalidSessionIdException`. Out of scope, not forced.
- **`NoSuchDriverException`** is "raised when driver is not specified and cannot be located" -- about locating the driver binary, not a dead session. Deliberately excluded.
- **No single chokepoint** exists in the Appium driver: each public method calls `self.driver.<x>` independently. The masking does converge in `strategies.py`, which is the effective chokepoint for every reported symptom.

Verified against the installed clients: Appium-Python-Client 5.3.1, selenium 4.44.0. The exception is `selenium.common.exceptions.InvalidSessionIdException`; the Appium client does **not** re-export it.

## Out of scope (deliberate)

No automatic reconnection, no retry loop that transparently re-establishes a session and resumes the keyword. This PR is "detect clearly and fail with one honest error". The recovery feature is larger and the audit itself treats it as optional/future work. The E0106 message names the recovery action accordingly ("terminate this session and start a new one") -- accurate, since nothing in the framework re-attaches on its own.

## Testing

`tests/units/common/test_strategies.py` -- new `TestDeadSessionIsNotMasked`: each masking site now yields `E0106` instead of its old code; the case where the element source already re-wrapped it as `E0201` with `cause` set; a guard that `E0106` stays outside the `E02*`/`X0201` family; and a negative case keeping an ordinary `WebDriverException` at `E0201`.

`tests/units/engines/drivers/test_appium_session_liveness.py` *(new)* -- the probe's three outcomes, plus `launch_app` calling `start_session` for a stale cached session and leaving a live one alone.

Full suite green (`pytest -c pytest.toml`, exit 0). `pre-commit run --files <changed>` clean (ruff, bandit, gitleaks, whitespace).

### Noted in passing, not fixed here

`StrategyManager._alloc_time_for_strategy` floors the first allocation to 0 when `timeout=1` and two or more strategies apply, breaking out before running a single strategy -- an `assert_presence` with a 1s budget silently asserts nothing. Pre-existing and unrelated to this PR; happy to open an issue.

## Update (post-review)

Detection is now a **per-driver contract** rather than WebDriver baked into the common layer. `DriverInterface.is_dead_session_error(exc)` is a generic seam (default `False`), and `common/session_liveness.py` no longer imports selenium: it consults whichever `DriverInterface` subclasses are loaded and walks the cause/context chain.

This supersedes two claims above:

- **Selenium** is no longer "left alone" for detection: it overrides the hook with the same `InvalidSessionIdException` signal, so it has explicit parity (its `launch_app` false-positive is still not shared, as described above).
- **Playwright** is no longer out of scope: it overrides the hook to map a closed browser/context/page (matched on the `Error` message, since the pinned version keeps `TargetClosedError` off the public API).

New backends now opt in by overriding one method instead of editing the common layer. Added `tests/units/engines/drivers/test_driver_dead_session_detection.py` pinning each backend's signal and the default. What stays out of scope is unchanged: no automatic reconnection.


optics_framework/common/session_liveness.py
```diff
@@ -0,0 +1,65 @@
+"""Recognition of a driver session that the remote server has dropped.
+
+An Appium/Selenium hub can terminate a session server-side -- idle reaping, a hub
+restart, a device being reclaimed -- while the client keeps its cached
```
malto101: making it very specific to Webdriver

make it generic 
if you want, add these implementation within each drivers

tests/units/common/test_strategies.py
```diff
@@ -625,3 +625,99 @@ def test_no_capable_strategy_raises_rather_than_returning_empty(self, monkeypatc
         with pytest.raises(OpticsError) as exc_info:
             manager.get_interactive_elements()
         assert exc_info.value.code == Code.E0202
+
+
+# --- Dead driver session (E0106) is not masked as element/pagesource not-found ---
+
+
+class TestDeadSessionIsNotMasked:
```
malto101, line 632: when adding new functionality, dont mention about past misatakes as such, as new people won't have idea about it

malto101 (commented):
if you are adding for appium
do add support for others too

other we will need maintain another doc page for feature parity

optics_framework/common/session_liveness.py
chinmayajha: Done. Pulled selenium out of the common layer: each driver now owns its own signal through `DriverInterface.is_dead_session_error` (default False, so a backend with no server-reaped session opts out), and `session_liveness` just asks whichever drivers are loaded and walks the cause/context chain. Appium and Selenium match `InvalidSessionIdException`, Playwright matches a closed browser/context/page. New backends override one method instead of touching common.

tests/units/common/test_strategies.py
chinmayajha: Reworded, dropped the backstory. It just states the invariant now: a dropped session is a driver-level failure, so it surfaces E0106 instead of an element-not-found result. No reference to how the handlers behaved before.

chinmayajha:
Pushed 3 commits addressing the review.

Made dead-session detection a per-driver contract instead of baking WebDriver into the common layer. `DriverInterface.is_dead_session_error(exc)` is the seam (default False), and Appium, Selenium and Playwright each override it with their own signal, so there's genuine parity across drivers and no separate feature-parity doc to maintain. 

`session_liveness` no longer imports selenium; it consults whichever driver subclasses are loaded. Selenium shares Appium's `InvalidSessionIdException`; Playwright maps a closed browser/context/page. A new cross-driver test pins each backend's signal.

Also reworded the strategies test docstring to state the invariant rather than the old handler behavior, and updated the E0106 / driver-backend notes in CLAUDE.md. Full unit suite green. @malto101

malto101:
@chinmayajha could you fix those sonar qube issues

chinmayajha:
> @chinmayajha could you fix those sonar qube issues

Missed them, pushed the new changes. @malto101

thakur-patel (approved):
LGTM

PR #517, opened 2026-08-31
fix: reword post-install message so it stops implying a running server

### LINKED ISSUE #494: Reword "Engines installed successfully!" — it overstates what happened
After answering yes to "Install Appium now?" during \`optics quickstart\`, the message "Engines installed successfully!" only means the Python \`appium-python-client\` package was pip-installed — not that the real Appium server (a separate Node.js process) is running.

The very next screen (the auto-run \`doctor\` table) contradicts it with "appium server: not reachable," right after being told installation "succeeded" — confusing for a newcomer who doesn't yet know these are two different things.

**Ask:** something like "Python client installed — the Appium server itself still needs to be running" would remove the contradiction.

Found during a from-scratch onboarding audit (optics-framework 1.10.0).

chinmayajha:
Closes #494

## Problem

`install_extras` (`optics_framework/helper/setup.py`) returned `"Engines installed successfully!"` after a successful pip install. For the mobile path that install is just `appium-python-client` — the Appium server is a separate Node process that is neither started nor checked here. During `optics quickstart` the next screen (auto-run `doctor`) immediately said `appium server: not reachable`, directly contradicting the "success" message a newcomer had just read.

## Change

One string, now:

> Engine packages installed. Services like the Appium server still need to be started separately.

`install_extras` is generic across every engine extra (Appium, Selenium, Playwright, OCR, LLM), so the wording stays generic and claims only what actually happened — packages installed — while naming the service users are most likely to trip over. It stays a single line, which matters because the setup TUI shows `message.splitlines()[0]` in its notification.

All three consumers pick the new text up unchanged: `optics setup --install` (`cli.py`), the setup TUI status line (`setup.py:_on_install_finished`), and `optics quickstart` (`quickstart.py:_offer_engine_install`).

## Tests

Updated the assertions in `tests/units/helpers/test_setup.py` and `tests/units/helpers/test_quickstart.py` that pinned the old wording (one exact-equality check plus several substring checks and mocked return values). The exact-equality check now carries a short note on why the wording matters, so a future reword doesn't silently reintroduce the overstatement.

`pytest tests/units/helpers/test_setup.py tests/units/helpers/test_quickstart.py` — 88 passed. `pre-commit` clean on the three changed files.

### DIFF
--- a/optics_framework/helper/setup.py
+++ b/optics_framework/helper/setup.py
@@ -471,7 +471,8 @@ def install_extras(requests: List[InstallRequest], *, stream: bool = True) -> tu
             # pulls the required OS libraries.
             run([sys.executable, "-m", "playwright", "install", "--with-deps", "chromium"])
 
-        return True, "Engines installed successfully!"
+        return True, ("Engine packages installed. Services like the Appium "
+                      "server still need to be started separately.")
     except subprocess.CalledProcessError as e:
         # capture_output / _run_streaming route the command's diagnostics to
         # e.stderr/e.stdout rather than losing them, so fold them into the
--- a/tests/units/helpers/test_quickstart.py
+++ b/tests/units/helpers/test_quickstart.py
@@ -91,7 +91,7 @@ class Harness:
 
     def __init__(self, ask_values, confirm_values, *,
                  answers=None,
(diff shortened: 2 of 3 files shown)


thakur-patel (approved):
LGTM

PR #522, opened 2026-08-31
fix: preserve separators in screenshot filename sanitizer

### LINKED ISSUE #521: Screenshot filename sanitizer silently strips characters instead of preserving separators
## Background

While fixing #500 (PR #520), we found that `action_keyword.py`'s `_save_screenshot_if_available(screenshot_np, f"pre-{func_name}")` call produced `prepress_element.jpg` instead of the intended `pre-press_element.jpg`. The call site already used a hyphen correctly — the hyphen was being silently stripped downstream by the filename sanitizer in `save_screenshot`:

https://github.com/mozarkai/optics-framework/blob/main/optics_framework/common/utils.py#L354

```python
name = re.sub(r'[^a-zA-Z0-9\s_]', '', name)
```

PR #520 fixed the one broken call site by switching to `f"pre_{func_name}"` (underscore is in the allow-list) rather than widening the shared sanitizer, since loosening it would change filenames for every other caller of `save_screenshot`. That was the right minimal fix for #500, but it routes around the underlying problem rather than closing it.

## The underlying problem

The sanitizer's allow-list (`[a-zA-Z0-9\s_]`) silently **deletes** any other character (hyphens, dots, colons, parentheses, etc.) instead of either:
- rejecting/erasing them in a way that's visible (e.g. replacing with `_` so word boundaries survive), or
- widening the allow-list to include common, filesystem-safe separator characters like `-` and `.`.

Deleting silently means two different intended names can collide into the same string (e.g. `pre-press_element` and `prepress_element` both sanitize to `prepress_element`), and there's no warning or log line when this happens — it's a silent correctness bug waiting for the next call site that uses a hyphen, dot, or other separator in a dynamic name.

## Duplicated copy of the same regex

`optics_framework/helper/live.py` (`LiveController.capture_screenshot`, around line 1236) independently reimplements the exact same sanitizer to predict the filename `save_screenshot` will produce, so it can return the correct path to the caller:

```python
sanitized = re.sub(r"[^a-zA-Z0-9\s_]", "", name)
```

This works today only because `name = "live_capture"` never contains a disallowed character. If `utils.save_screenshot`'s sanitizer regex is ever changed (e.g. to fix this issue) without updating this duplicate, `optics live`'s `/screenshot` command will silently return a path that doesn't match the actual file on disk.

## Suggested fix

1. Change the sanitizer in `utils.save_screenshot` to replace disallowed characters with `_` instead of deleting them (`re.sub(r'[^a-zA-Z0-9\s_]', '_', name)`), so separators are preserved as word boundaries and two different names are far less likely to collide.
2. Extract the sanitizer into a small shared helper (e.g. `utils.sanitize_screenshot_name(name)`) and have both `utils.save_screenshot` and `LiveController.capture_screenshot` call it, instead of maintaining two copies of the same regex.
3. Audit other dynamic screenshot-name call sites (grepped as part of #500's fix — see `grep -rn "save_screenshot(" optics_framework/`) to confirm none currently rely on the old deletion behavior in a way that a replace-with-underscore change would visibly break (as of this filing, all other call sites use static underscore-only names, so this should be low risk).

## Blast radius already checked

- Grepped all `save_screenshot(...)` call sites (`test_runnner.py`, `pytesseract.py`, `remote_ocr.py`, `action_keyword.py`, `verifier.py`, `remote_oir.py`, `easyocr.py`, `live.py`) — only `action_keyword.py:433` (fixed by #520) and `live.py`'s dynamic `name` param currently pass anything through the sanitizer; every other call site uses a static, already-safe string.
- No test currently pins the exact deletion (vs. replacement) behavior of the sanitizer regex itself, so a replace-with-underscore change should be a small, low-risk, well-testable follow-up.

Low priority / good-first-issue — not urgent, since no currently-shipped call site is actively broken after #520.

chinmayajha:
Closes #521

## The bug

`utils.save_screenshot` sanitized the caller-supplied name by **deleting** every character outside its allow-list:

```python
name = re.sub(r'[^a-zA-Z0-9\s_]', '', name)   # before
```

Deletion silently merges names that differ only by a separator. Verified before touching anything:

```
'pre-press_element'  -> 'prepress_element'
'prepress_element'   -> 'prepress_element'   # same file, no warning
```

That is exactly how #500 produced `prepress_element.jpg` from a call site that correctly passed `pre-{func_name}`.

`LiveController.capture_screenshot` (`helper/live.py`) independently runs the same regex, purely to predict the filename `save_screenshot` is about to write so it can return the path to `/screenshot`.

## The fix

Both copies get the identical one-character fix — `''` → `'_'` — so disallowed characters are **replaced**, not deleted:

```python
name = re.sub(r'[^a-zA-Z0-9\s_]', '_', name)   # after
```

| | before | after |
|---|---|---|
| `pre-press_element` | `prepress_element` | `pre_press_element` |

Minimal, 2-file, 2-line diff — no new function, no behavior change beyond fixing the delete-vs-replace bug.

**Known tradeoff, left as-is deliberately**: the regex still lives in two places (`utils.py` and `live.py`). An earlier version of this PR extracted a shared `sanitize_screenshot_name()` helper so the two couldn't drift again; that was reverted in favor of keeping this fix as small as possible. If the pattern changes again in the future, both call sites need updating together — `grep -rn "a-zA-Z0-9\\\\s_" optics_framework/` finds both.

## Call-site audit (re-verified, not taken from the issue)

`grep -rn "save_screenshot(" optics_framework/` — every call site was diffed old-vs-new:

- **Unchanged, byte for byte** (all already underscore-only): `action_keyword.py` `press_keycode`, `execute_script`, `enter_text_keyboard`, `enter_text_using_keyboard`, `{func_name}_with_aoi`; `verifier.py` `capture_screenshot`; `test_runnner.py` `end_of_run`; `remote_oir.py` `detected_template`; `remote_ocr.py` `detected_text`; `pytesseract.py` `annotated_frame`; `easyocr.py`; `live.py` `live_capture`.
- **Changed (intentionally, to the correct value)**: `action_keyword.py:433` `f"pre-{func_name}"` → now writes `pre_press_element.jpg` instead of `prepress_element.jpg`.

**Interaction with #520** (open, `docs/500-cli-polish-papercuts`): that PR routes around this by changing the call site to `f"pre_{func_name}"`. Both orders converge on the same filename `pre_press_element.jpg`, and this PR does not touch `action_keyword.py`, so there is no conflict and no merge-order dependency either way.

No test in `tests/` pinned the old deletion behaviour. (`test_live_driver_agnostic.py:333` writes a literal `pre-press_element.jpg` straight to disk as a teardown fixture — it never goes through the sanitizer.)

## Judgment call: consecutive disallowed characters are NOT collapsed

`a..b` becomes `a__b`, not `a_b`. Deliberate:

- Collapsing reintroduces the exact bug class being fixed — it makes the mapping non-injective again (`a..b` and `a_b` would collide).
- A 1:1 character mapping means two distinct inputs of equal length can never produce the same output.
- It matches existing repo precedent: `expose_api.py:673` already does `re.sub(r"[^a-zA-Z0-9_.-]", "_", name)` without collapsing.

## Tests

No automated test added — verified manually instead:

```pycon
>>> import re
>>> s = lambda n: re.sub(r'[^a-zA-Z0-9\s_]', '_', n)
>>> s("pre-press_element"), s("prepress_element")
('pre_press_element', 'prepress_element')
```

The two names the issue was filed over no longer collide.

**Full suite**: `pytest` → 1062 passed, 2 xfailed, 1 failed. The 1 failure is
`TestExtractionCost::test_cost_growth_is_linear_not_quadratic`, a known load-sensitive flake that fails only
under parallel load and passes when run alone (verified) — unrelated to this change (no timing code touched).

`pre-commit run --files <changed>`: ruff, bandit, trailing-whitespace, end-of-file-fixer, gitleaks all pass.

### DIFF
--- a/optics_framework/common/utils.py
+++ b/optics_framework/common/utils.py
@@ -344,7 +344,7 @@ def save_screenshot(img, name, output_dir, time_stamp=None):
     if output_dir is None:
         internal_logger.info(OUTPUT_PATH_NOT_SET_MSG)
         return
-    name = re.sub(r'[^a-zA-Z0-9\s_]', '', name)
+    name = re.sub(r'[^a-zA-Z0-9\s_]', '_', name)
     if time_stamp is None:
         time_stamp = str(datetime.now().astimezone().strftime('%Y-%m-%dT%H-%M-%S-%f'))
     screenshot_file_path = os.path.join(output_dir, f"{time_stamp}-{name}.jpg")
--- a/optics_framework/helper/live.py
+++ b/optics_framework/helper/live.py
@@ -1235,7 +1235,7 @@ class LiveController:
         name = "live_capture"
         timestamp = datetime.now().astimezone().strftime("%Y-%m-%dT%H-%M-%S-%f")
         utils.save_screenshot(image, name, output_dir=output_dir, time_stamp=timestamp)
-        sanitized = re.sub(r"[^a-zA-Z0-9\s_]", "", name)
+        sanitized = re.sub(r"[^a-zA-Z0-9\s_]", "_", name)
         return os.path.join(output_dir, f"{timestamp}-{sanitized}.jpg")
 
     def screenshot_png_bytes(self) -> bytes:


malto101:
dont see a reason to merge this
we are following a semantic, why should we allow "-" in function name?

chinmayajha:
> dont see a reason to merge this we are following a semantic, why should we allow "-" in function name?

I understand the semantic, but then why do we have a hyphen at the call site, it should be consistent right? There's no point in providing a hyphen at the call site and filtering it later by regex.

two solutions
1. Either don't filter, differentiate (this PR)
2. Keep filter, fix call site (I can close this PR and open that if needed) @malto101

malto101:
> two solutions
> 1. Either don't filter, differentiate (this PR)
> 2. Keep filter, fix call site (I can close this PR and open that if needed) @malto101 

I would suggest to fix call site, as it wouldn't make sense to accept both hypen and underscores 

Let's keep a simple semantic so it will be easier for end user to remember function/names and reduce the cognitive complexity around it

chinmayajha:
> I would suggest to fix call site, as it wouldn't make sense to accept both hypen and underscores
> 
> Let's keep a simple semantic so it will be easier for end user to remember function/names and reduce the cognitive complexity around it

okay, will close this PR once the other one is up. thank you!

chinmayajha:
Per the review discussion (keep the `[a-zA-Z0-9\s_]` semantic, fix the call site), this PR is superseded by #525 — which fixes the call site (`pre-{func_name}` → `pre_{func_name}` in `action_keyword.py`) after auditing every `save_screenshot` call site in the tree. Closing this one as agreed.

chinmayajha:
Closing in favor of #525 (call-site fix per review decision).

PR #530, opened 2026-09-04
fix: sync strict_element_match through the HTTP/MCP API and web drivers

chinmayajha:
## Summary

This PR fixes `strict_element_match` not being passed through `optics serve` and `optics mcp`.

Previously, the flag was supported by Appium but was dropped by the API/session layer. It is now exposed through `SessionConfig` and the MCP `start_session` tool.

Strict matching is also now supported by Selenium and Playwright:

* Selenium skips partial/`contains()` matching when strict mode is enabled.
* Playwright uses `exact=True` in strict mode.
* Presence/assertion checks always use exact matching.


thakur-patel:
@chinmayajha does this work for all drivers?

chinmayajha:
> does this work for all drivers?

it does now as this is not a driver specific behaviour (config flag applies to all). updated the PR now

thakur-patel (approved):
LGTM

PR #537, opened 2026-09-07
fix(appium): activate the app when launch_app runs on a live session

ARYANK-08:
### Fix: Launch App now launches on a live session

Previously, `launch_app` only worked when no driver existed. Since session creation calls it with empty parameters, subsequent Launch App actions returned the session ID without opening the app.

This caused ** Launch App** silently no-op.

**Fix:** When a session is already active and an app identifier is provided, `launch_app` now activates the app on the live driver using `launch_other_app`. Calls without an identifier retain the existing behavior, so session creation is unaffected.

### DIFF
--- a/optics_framework/engines/drivers/appium.py
+++ b/optics_framework/engines/drivers/appium.py
@@ -707,7 +707,7 @@ class Appium(DriverInterface):
         app_activity: Optional[str] = None,
         event_name: Optional[str] = None,
     ) -> str:
-        """Launch the app using the Appium driver."""
+        """Launch the app, activating it on the driver if a session is already alive."""
         session_id = self.get_session_id()
         if self.driver is None or not self._is_session_alive():
             session_id = self.start_session(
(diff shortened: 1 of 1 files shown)


optics_framework/engines/drivers/appium.py
```diff
@@ -707,14 +707,16 @@ def launch_app(
...
+        """Launch the app, activating it on the driver if a session is already alive."""
         session_id = self.get_session_id()
         if self.driver is None or not self._is_session_alive():
             session_id = self.start_session(
                 app_package=app_identifier,
                 app_activity=app_activity,
                 event_name=event_name,
             )
+        elif app_identifier:
+            self.launch_other_app(app_identifier, event_name)
```
kun-codes: Silently using `launch_other_app` when `launch_app` is called is bad practice in my opinion

optics_framework/engines/drivers/appium.py
```diff
@@ -707,14 +707,16 @@ def launch_app(
         app_activity: Optional[str] = None,
```
kun-codes, line 707: nothing wrong with this line but update the docstring at https://github.com/mozarkai/optics-framework/blob/7d5554865069cde0539688711695e4bbec61b621/optics_framework/api/app_management.py#L37

chinmayajha (approved):

PR #541, opened 2026-09-09
fix: skip local adb device check for remote Appium servers

chinmayajha:
## Summary
- The `optics execute` preflight had a mismatch which checked for local adb + at least 1 device even when remote appium hub URL was being used. 
- Added `_is_local_host` to detect whether the Appium URL resolves to this machine (loopback), and only run the local `adb` device check in that case.

### DIFF
--- a/docs/architecture/cli_layer.md
+++ b/docs/architecture/cli_layer.md
@@ -126,7 +126,7 @@ sequenceDiagram
     CLI-->>User: Results displayed (+ failure details when steps failed)
 ```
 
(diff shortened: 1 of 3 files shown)


malto101:
Wouldn't it be better to just not have local adb runs in the whole repo?, we could you APPIUM/selenium remote shell execution to help us with adb discovery?

chinmayajha:
Kept as-is: preflight runs before any session exists (execute.py:913), so mobile: shell can't help with discovery there, and this PR already stops local adb for any non-loopback Appium URL (execute.py:632), which was the actual bug. The other local adb call sites (doctor advisory checks, live device picker/hotplug) only make sense on this machine since they enumerate USB-attached devices. Happy to file a follow-up issue if you want Appium-side device discovery for remote hubs.

PR #545, opened 2026-09-09
fix(app_management): pass event_name as keyword to launch_app (#538)

### LINKED ISSUE #538: AppManagement.start_appium_session passes event_name into launch_app's app_identifier slot
## Summary

`AppManagement.start_appium_session` calls the driver's `launch_app` positionally:

```python
# optics_framework/api/app_management.py:51-57
def start_appium_session(self, event_name: Optional[str] = None) -> None:
    """
    Starts an Appium session.

    :param event_name: The event triggering the session start, if any.
    """
    self.driver.launch_app(event_name)
```

`Appium.launch_app`'s signature is `launch_app(self, app_identifier=None, app_activity=None, event_name=None)`, so the `event_name` string lands in the **`app_identifier`** parameter, not `event_name`.

## Why this is currently invisible

On `main`, `launch_app` only does anything when no session exists yet (`self.driver is None or not self._is_session_alive()`), in which case it calls `start_session(app_package=app_identifier, ...)`. `start_session` only uses `app_package` to *override capabilities* for a brand-new session — a stray event-name string there is harmless noise in the capabilities dict, so the bug produces no observable failure today.

## Why it's about to become live

[PR #537](https://github.com/mozarkai/optics-framework/pull/537) makes `launch_app` do something when a session is **already alive**:

```python
elif app_identifier:
    self.launch_other_app(app_identifier, event_name)
```

`launch_other_app` calls `driver.activate_app(app_name)`. Once #537 merges, calling `Start Appium Session` with a non-empty `event_name` against an already-alive session will pass that event-name string to `activate_app(...)`, which is not a valid package/bundle id, and raise:

```
OpticsError: Failed to launch: <event_name>
```

So a keyword call that is silently a no-op today will start throwing after #537, for any suite where `Start Appium Session` runs more than once against a session that's already up (e.g. a second test case in the same `optics execute` batch run).

## Where this is reachable

- `Optics.start_appium_session` (`optics_framework/optics.py`) — the public SDK/Robot keyword.
- `optics generate` emits `"Start Appium Session"` (`optics_framework/helper/generate.py:141`) as part of generated pytest/Robot code, so this isn't limited to hand-written CSV/YAML suites.

## Existing test encodes the bug as correct behavior

`tests/units/api/test_app_management.py:31`:

```python
("start_appium_session", ("ev",), "launch_app", ("ev",), {}),
```

This asserts `start_appium_session("ev")` delegates to `driver.launch_app("ev")` — i.e. it pins the current (wrong) positional mapping as expected, using a mock driver that doesn't exercise the real `Appium.launch_app` parameter binding.

## Suggested fix

```python
def start_appium_session(self, event_name: Optional[str] = None) -> None:
    self.driver.launch_app(event_name=event_name)
```

...and update the parametrized case in `test_app_management.py` to assert the corrected keyword-argument call shape.

## Suggested priority

Should land before or alongside #537, since #537 is what converts this from a dormant bug into a real runtime failure for existing "Start Appium Session" usages (including generated suites).

aeroscissorz:
## Summary

AppManagement.start_appium_session was calling the driver's `launch_app` positionally:

`python
self.driver.launch_app(event_name)
`

Since `launch_app`'s signature is `launch_app(self, app_identifier=None, app_activity=None, event_name=None)`, the `event_name` string was being bound to `app_identifier` instead of `event_name`.

## Why it matters

Today this is a silent no-op: `launch_app` only acts when no session exists, where a stray string in `app_identifier` is harmless noise in the capabilities dict.

Once #537 merges, `launch_app` will act on an already-alive session:

`python
elif app_identifier:
    self.launch_other_app(app_identifier, event_name)
`

That forwards the event-name string to `activate_app(...)`, which is not a valid package/bundle id, raising:

`
OpticsError: Failed to launch: <event_name>
`

This would break any suite where **Start Appium Session** runs more than once against a live session (e.g. a second test case in the same `optics execute` batch run, and generated pytest/Robot suites via `optics generate`).

## Fix

`python
def start_appium_session(self, event_name: Optional[str] = None) -> None:
    self.driver.launch_app(event_name=event_name)
`

Updated the parametrized case in `tests/units/api/test_app_management.py` to assert the corrected keyword-argument call shape.

## Testing

`
python -m pytest tests/units/api/test_app_management.py -v
# 11 passed
`

Fixes #538

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ optics.setup(
     elements_sources=[{"appium_find_element": {"enabled": True}}],
 )
 
-optics.launch_app("com.example.app")
+optics.launch_app(app_identifier="com.example.app")
 optics.enter_text("username_field", "testuser")
 optics.press_element("submit_button")
 optics.validate_element("welcome_message")
--- a/docs/architecture/components.md
+++ b/docs/architecture/components.md
@@ -205,7 +205,7 @@ enter_text(["username", "user_name", "//input[@name='user']"], "testuser")
 **3. Environment-Specific Values:**
 ```python
 # Different values for different environments
-launch_app([
+launch_app(app_identifier=[
     "com.example.app.dev",      # Development
     "com.example.app.staging",  # Staging
     "com.example.app.prod"      # Production
--- a/docs/architecture/library_layer.md
+++ b/docs/architecture/library_layer.md
@@ -97,7 +97,7 @@ optics.setup(
 )
 
 # Use keywords
-optics.launch_app("com.example.app")
+optics.launch_app(app_identifier="com.example.app")
 optics.press_element("submit_button")
 optics.enter_text("username_field", "testuser")
 optics.validate_element("welcome_message")
@@ -175,7 +175,7 @@ optics.scroll(100, 200, 300, 400)
 
 ```python
 # App lifecycle
-optics.launch_app("com.example.app")
+optics.launch_app(app_identifier="com.example.app")
 optics.launch_other_app("com.other.app")
 optics.close_and_terminate_app()
 optics.force_terminate_app()
@@ -250,8 +250,8 @@ optics1 = Optics(config=config1)
 optics2 = Optics(config=config2)
 
 # Each instance has its own session
-optics1.launch_app("app1")
-optics2.launch_app("app2")
+optics1.launch_app(app_identifier="app1")
+optics2.launch_app(app_identifier="app2")
 ```
 
 ## Configuration
--- a/docs/usage/library_usage.md
+++ b/docs/usage/library_usage.md
@@ -27,7 +27,7 @@ optics.setup(
 )
 
 # Launch app
-optics.launch_app("com.example.app")
+optics.launch_app(app_identifier="com.example.app")
 
 # Interact with elements
 optics.press_element("submit_button")
@@ -134,7 +134,7 @@ optics.setup(config=yaml_config)
 
 ```python
 # Launch app by package name
-optics.launch_app("com.example.app")
+optics.launch_app(app_identifier="com.example.app")
 
 # Launch other app
 optics.launch_other_app("com.other.app")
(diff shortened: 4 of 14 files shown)


chinmayajha:
Thanks for this. Can you also check other areas of the repository where a similar issue will occur? i.e. positional arguments issue basically. @aeroscissorz

aeroscissorz:
Sure

aeroscissorz:
I checked the repo and #538 turned out to be the only occurrence. :)

chinmayajha:
> I checked the repo and #538 turned out to be the only occurrence. :)

https://github.com/search?q=repo%3Amozarkai%2Foptics-framework+launch_app%28&type=code

Check this URL, there are several instances. @aeroscissorz

aeroscissorz:
Pls correct me if im wrong, those are definitions, docs, tests, and correct keyword calls. the sole defective call is the one this PR fixes, and it only still shows in search because the search targets un-merged .main. every actual invocation either uses keyword arguments or places arguments in the correct positional slots.

chinmayajha:
> Pls correct me if im wrong, those are definitions, docs, tests, and correct keyword calls. the sole defective call is the one this PR fixes, and it only still shows in search because the search targets un-merged .main. every actual invocation either uses keyword arguments or places arguments in the correct positional slots.

I understand that. I'm just saying that prior to the issue #538 there was no convention of ensuring positions, the places just mentioned whatever they intended to (app_activity, event_name, whatever was convenient)

But now that we have to follow, let's have it uniformly. So basically every mention in the repository would now need to have `launch_app(event_name=event_name)` instead of `launch_app(event_name)` and so on, is what I'm saying. @aeroscissorz 

Once this is done, this PR would have handled all surfaces and will get merged soon :)

chinmayajha:
Thanks @aeroscissorz 
Please squash your commits into 1. I shall approve it after that.

chinmayajha:
Thank you for your contribution @aeroscissorz

chinmayajha (approved):

PR #549, opened 2026-09-11
fix: make presence assertions honour strict_element_match, uniform with locate

chinmayajha:
## What

This routes the `assert presence` path through the same `_strict_element_match()` flag as `locate`, so the two stay uniform:

## Behaviour change (intentional)

With the flag at its **`False` default, presence checks now fuzzy-match by default.**


thakur-patel (approved):
LGTM

PR #551, opened 2026-09-11
fix(runner): stop variable-bearing yaml steps dispatching to a shorter keyword

chinmayajha:
## Problem

#547 resolves a YAML step's keyword from the keyword catalogue before splitting on `${...}`. A misspelled multi-word keyword whose prefix is itself a keyword now dispatches to the shorter keyword, losing the unknown-keyword error and its "did you mean" hint:

| YAML step | before #547 | main today |
|---|---|---|
| `Swipe By Percent ${x} ${y}` | `Swipe By Percent` -> *Did you mean 'Swipe By Percentage'?* | `Swipe` + `["By","Percent",...]` |
| `Enter Text Using Keybord ${f} hi` | `Enter Text Using Keybord` -> suggestion | `Enter Text` + `["Using","Keybord",...]` |
| `Press Element With Indx ${el} 2` | `Press Element With Indx` -> suggestion | `Press Element` + `["With","Indx",...]` |

`optics dry_run` reports PASS for these typos, and the runtime failure becomes a param/type error instead of the hint. #548 also unwraps every fully-quoted value (`text="hello"` -> `text=hello`), changing params that previously survived verbatim.

## Fix

- Variable-bearing steps keep the legacy split. If a step contains `${`, the keyword is the text before the first `${`, exactly as before #547. Catalogue resolution is used only for fully literal steps, which keeps #547's fix (`Swipe 1000 300 up`, `Sleep 5`, slug forms) and #548's quoted-space params (`text="two words"`).
- Quotes are preserved unless they carry a space: a fully quoted value is unwrapped only when its body holds whitespace. `text="hello"`, `key="a=b"` and `"//button"` read as they did on 1.10.3.
- Module references match case/whitespace-insensitively on the whole step and return the canonical project name. Names containing `${` are skipped so they cannot shadow the variable split.
- `read_module_names` ignores malformed `Modules` shapes (scalar, mapping, list of non-mappings) instead of leaking characters.
- Third-party readers keep loading: `_load_modules` uses `read_module_names` when available, falls back to a full read otherwise, and only passes the names argument when the reader declares `module_names`.
- The step parser was split into small helpers to clear SonarQube `python:S3776` (cognitive complexity 16 > 15).

## Compatibility

- CSV parsing is byte-identical.
- For every step containing a complete `${...}` variable, the parse matches the pre-#547 parser except the intentional #548 quoted-space values.
- The only outputs differing from both trees are `Press Element "//button"` and `Sleep "Well"` (quoted no-space param on a now-resolved keyword); neither parsed on 1.10.3.
- Known limitation (unchanged from #547): a fully literal typo whose prefix is a keyword, e.g. `Swipe By Percent 50 50 20`, still binds to `Swipe`; dry-run arity validation could be a follow-up.
- `read_modules` still raises on malformed `Modules` (pre-existing).

## Verification

- `pytest tests/units` -> 1170 passed, 2 xfailed; `ruff check` clean.
- Parsed a 639-case YAML step corpus against the pre-#547 tree, `main`, and this branch: only the intended cases differ (typo names restored, whitespace-carrying quotes unwrapped, module references canonicalised).

Refs #547, #548.

### DIFF
--- a/optics_framework/common/runner/data_reader.py
+++ b/optics_framework/common/runner/data_reader.py
@@ -35,6 +35,10 @@ def _tokens(step: str) -> List[str]:
     """Whitespace-split, except that a quoted run holds together and a value written entirely
     in quotes is unwrapped — which is the only way a param can contain a space.
 
+    Unwrapped whatever the value holds, not only when it holds a space: an editor writing
+    every param as `name="value"` would otherwise hand the keyword the quotes as well, and
+    `index="2"` would arrive as the string `"2"`.
+
     An unbalanced quote falls back to `.split()`: the pattern would drop the quote and split
     anyway, so `text=it's fine` keeps reading as it always has."""
     if _unbalanced_quote(step):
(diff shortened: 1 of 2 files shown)


kun-codes:
@chinmayajha  I have pushed some changes, can you review them? Thanks

kun-codes (approved):

PR #552, opened 2026-09-14
fix(strategies): clamp vision assert polling to the allocated slice

### LINKED ISSUE #524: TextDetectionStrategy/ImageDetectionStrategy.assert_elements overshoot their allocated slice by up to ~3x
## What

`StrategyManager._alloc_time_for_strategy` (fixed in #518) guarantees each applicable strategy gets a non-zero, honest slice of the caller's `assert_presence`/`assert_visibility` timeout. But `TextDetectionStrategy.assert_elements` and `ImageDetectionStrategy.assert_elements` (`optics_framework/common/strategies.py`) don't actually honor that slice tightly — both poll in fixed `self.screenshot_timeout = 1.5s` chunks regardless of how small their allocation is:

```python
def assert_elements(self, elements, timeout=30, rule='any'):
    end_time = time.time() + timeout
    ...
    while time.time() < end_time:
        time.sleep(self.screenshot_timeout)   # 1.5s, unconditional
        frames = ss_stream.get_all_available_screenshots(wait_time=1)
        if not frames:
            time.sleep(self.screenshot_timeout)   # another 1.5s if nothing arrived yet
            continue
        ...
```

Given a 0.5s allocation (the new floor from #518), the strategy still burns at least 1.5s, and up to ~3s if no screenshot is available on the first pass — before it even checks `end_time` again.

## Evidence (live device, via #518's own verification)

Ran `StrategyManager.assert_presence(["text"], "Text", timeout=1)` against a real Android device through the Appium hub, with `TextElementStrategy` + `TextDetectionStrategy` (pytesseract) as the two applicable strategies:

| timeout requested | actual elapsed |
|---|---|
| 1s | 2.60 - 2.89s |
| 3s | 5.21 - 5.72s |

The nominal per-strategy floor from #518 is respected (neither strategy gets skipped), but the real overshoot is 1.6-2.9s, not the ~0.5s estimated in #518's description — because `TextDetectionStrategy`'s own sleep granularity dominates once it's in the applicable set.

## Why this is separate from #518

`_alloc_time_for_strategy` only computes the *nominal* budget; it doesn't touch `TextDetectionStrategy`/`ImageDetectionStrategy.assert_elements`, which predate that fix and have their own timing behavior. Fixing #507 correctly stopped strategies from being silently skipped, but callers relying on `assert_presence(timeout=X)` finishing close to `X` when OCR/image strategies are in play will still see it run considerably longer.

## Suggested direction

`screenshot_timeout` (currently a fixed 1.5s) should probably be clamped to whatever's actually left of the strategy's own allocated slice, so a strategy given e.g. 0.5s doesn't unconditionally sleep past it. Needs a look at `capture_screenshot_stream`'s own timeout semantics too, since both strategies lean on it.

Discovered while live-verifying #518 against a real device (see that PR's discussion).

aeroscissorz:
## Summary

`TextDetectionStrategy.assert_elements` and `ImageDetectionStrategy.assert_elements` polled in fixed `1.5s` chunks regardless of their allocated slice, overshooting `assert_presence`/`assert_visibility` timeouts by up to ~3x (see #524).

## Changes

- Poll the screenshot stream first, then back off only when it's empty — so even a slice smaller than the poll interval (e.g. the `0.5s` floor) still inspects frames.
- Clamp the stream wait and the backoff sleep to whatever remains of the strategy's slice, so a `0.5s` slice can't bleed into a fixed `1.5s` sleep.
- The screenshot stream itself already runs for the strategy's `timeout`, so no change was needed there.

## Tests

- Updated `TestPollFramesWithinDeadline` and added a regression case proving a sub-poll-interval slice still polls and yields frames.
- `pytest tests/units/common/test_strategies.py` and `ruff check` pass.

Fixes #524


optics_framework/common/strategies.py
```diff
@@ -152,6 +152,44 @@ def _assert_elements_locator_style(
...
+    ) -> Generator[list, None, None]:
+        """Yield batches of available screenshots, never waiting past ``end_time``.
+
+        Polls the stream first so a strategy given a slice smaller than
+        ``poll_interval`` (e.g. the 0.5s floor from #518) still gets to inspect a
+        frame. When the stream has nothing ready it backs off for ``poll_interval``
+        -- clamped to the remaining budget -- before polling again. The stream wait
+        and the backoff are both bounded by ``end_time``, so a 0.5s slice can no
+        longer overshoot into a fixed 1.5s sleep (see issue #524).
+
```
malto101: ideally it would be nice to not have this deep doctoring explaination for an internal api whose implementation is quite direct and isnt obscure

optics_framework/common/screenshot_stream.py
malto101: as suggested by sonarqube, could you refactor this function to be lesser then 15

malto101 (changes requested):
rest all looks good,
I see no probs as such implementation wise

PR #559, opened 2026-09-16
fix(flow-control): create the element store an extract needs

kun-codes:
`Invoke API` cannot run in a session created by `optics serve`: its `extract` raises `Code.E0702: Session elements is not an ElementData instance or is None`, because `expose_api` creates every session with `elements=None` and nothing can fill it afterwards — the REST API has only a GET for elements, and `Add Element` is not a keyword `execute_keyword` registers.

`_get_or_create_session_elements` now creates the store its name promises. `TestRunner` already substitutes an empty `ElementData` in the same case, so this matches what the runner has always done.

Only the write path changed. The three reads that raise the same error (`_resolve_condition`, `_compute_expression`, `_evaluate_expression`) still do — no store there genuinely means no value.


optics_framework/api/flow_control.py
```diff
@@ -1289,11 +1289,18 @@ def _extract_and_store_single_value(self, response_data, element_name: str, path
...
+        """Gets session elements or creates if not properly initialized.
+
+        `optics serve` creates every session with `elements=None` and offers no way to fill
+        it, so raising here left `Invoke API` unusable there — its `extract` had nowhere to
+        put a value. `TestRunner` already substitutes an empty `ElementData` in the same
+        case, so creating one is the reading the runner has always taken.
+        """
         runner_elements = getattr(self.session, "elements", None)
         if not isinstance(runner_elements, ElementData):
-            raise OpticsError(Code.E0702, message=NO_SESSION_ELEMENT_PRESENT)
```
thakur-patel, line 1296: Was this error a false error previously?

chinmayajha (approved):
LGTM

